PR #451, opened 2021-06-10
fix: Alert display logic given Brookline Hills closure

losvedir:
#### Summary of changes
**Asana Ticket:** [🐛 report from PIOs about signs going blank instead of using shuttle text](https://app.asana.com/0/584764604969369/1200423798177911)

I reproduced the alert causing the bug and verified on signs-ui-dev the incorrect behavior, which looks like this:

Screen Shot 2021-06-10 at 12 39 52 PM

Basically, what's supposed to happen is that when there's an alert that covers a consecutive set of stops, all the middle stops should show "no train service" and the stops on the edges should be blank.

Realtime Signs does this by maintaining a configuration file `priv/stops.json` that lists the stops in order for each route, and checking if all its "neighboring" stops are also in the alert.

The issue in this case is that Brookline Hills is closed long-term and so it's been removed from GTFS and alerts-ui, and so its `70179` and `70178` stops aren't in the alert. Hence, the stops on both side of it think they're at an "edge" of the alert and go blank.

This is a simple PR that just updates the configuration to remove Brookline Hills. It's running on realtime-signs-dev and I verified on signs-ui-dev that with the recreated alert the signs do what they're supposed to.

However, this is not an ideal solution, since when Brookline Hills comes back we (well Screens) will have to remember to re-add it. I made a ticket, which we'll probably hand over with the app, that suggests changing the implementation from this hardcoded lists of stops to using an Alert's BOARD/EXIT/RIDE activities to infer the edges. This is what I learned Screens did for some similar functionality in their app, and feels more robust. So maybe they'll want to do that here, too. Alternatively, the GTFS canonical routes by help here by putting Brookline Hills back in alerts-ui and the alert IDs list.

#### Reviewer Checklist


mkennedm (approved):
Looks good.

PR #452, opened 2021-06-15
feat: latency analysis script

digitalcora:
[Rendered README](https://github.com/mbta/realtime_signs/blob/cfg-latency-analysis/scripts/latency_analysis/README.md)

**Asana:** [get raw data together for ARINC audit of Davis](https://app.asana.com/0/584764604969369/1200285366666696)

On the Asana task, I attached the current output of `elixir analyze.exs --filter sta=RDAV`, using the input files attached to the parent task.

### Notes

From the task specifications:

* _"Filter to just RDAVMSIGN002, RDAVNSIGN003, and RDAVSSIGN006, for a representative sign from the mezzanine, north, and south zones"_ — I just made it output stats for all signs, one CSV row per sign, reasoning it would be easy to filter as needed in Excel or whatever we'll use to analyze this. The **Output** section in the README explains it in more detail.

* _"MSGLOG: this is for **all** requests in the system, filter to just `sta=RDAV`"_  — I implemented this as a generic filter that can be provided on the command line. With the given input files it does take a _long_ time to process if `--filter sta=RDAV` is not used (it gets the same results, it just does a lot of extra work that is ultimately thrown away).

* _"If a request can't be matched to the MSGLOG log, drop it"_ — I left these in, since the patchwork nature of the logs means we often see an update on the SCU despite not seeing it on the head-end server, and so can generate partial stats for it. If these aren't wanted, they can be easily filtered out "in post".


bklebe (approved):
Looks good! Thanks Cora.

PR #465, opened 2021-11-02
feat: log process health every five minutes

bklebe:
#### Summary of changes
**Asana Ticket:** [Add logging to realtime_signs to find leak](https://app.asana.com/0/584764604969369/1200726999492021/f)

Added code for logging process health with Recon from [RTR](https://github.com/mbta/rtr/blob/c1e6152c6c901f989ee81a0177ffd307800972aa/rtr/lib/health/monitor.ex), as the ticket suggests. Cleaned up that code to use Streams instead of Lists to avoid multiple enumeration, added specs for it, and extracted all side effects into a top-level `Enum.each/2` in the message handler.

#### Reviewer Checklist


lib/engine/health.ex
```diff
@@ -14,18 +14,27 @@ defmodule Engine.Health do
...
 
+  @spec init(keyword) :: {:ok, Engine.Health.t()}
   def init(opts) do
     period_ms = Keyword.get(opts, :period_ms, @default_period_ms)
     network_check_mod = Keyword.get(opts, :network_check_mod, Engine.NetworkCheck.Hackney)
     restart_fn = Keyword.get(opts, :restart_fn, Application.get_env(:realtime_signs, :restart_fn))
 
     {:ok, timer_ref} = :timer.send_interval(period_ms, self(), :health_check)
 
+    Process.send_after(
```
losvedir, line 32: Since this blindly repeats on an interval, you can do `{:ok, _} = :timer.send_interval(@process_health_interval_ms, self(), :process_health)` like the line above, and then the `handle_info` for `:process_health` won't need to worry about sending itself a message later.

I'm not super sure why we're storing the `timer_ref` of the other interval. It doesn't look like we do anything with it, so you probably don't need to store the process health one.

losvedir:
This looks generally good to me. Can you deploy to realtime-signs-dev so we can let it run for a while and see it logging data?

lib/engine/health.ex
bklebe: My usage here was influenced by this part of the Erlang efficiency guide: https://www.erlang.org/doc/efficiency_guide/commoncaveats.html#timer-module. Maybe in this case we're not so worried about creating too many timers, or overloading this process with logging requests that it can't handle, since they're being sent infrequently?

losvedir: Yeah, I think in this case we can go with the simple `:timer.send_interval/3` approach.

losvedir (approved):
Looks good in the test environment. 👍 

🚀

PR #472, opened 2022-01-06
Set http updater max send rate per sec to 8

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Run rate-limiting experiment in Dev](https://app.asana.com/0/1199570165415679/1201622027899197/f)

This PR is to set the maximum send rate for the http updaters to 8, a 75% reduction from the original 32.

The purpose of this is to experiment to see if a slower rate of messages yields a wider gap in between latency spikes from the ARINC headend server. This change will **only be applied in dev** and will never in prod. My intention is to let the change bake in dev for a work day or so and observe how the latencies are affected from Splunk.

### DIFF
--- a/lib/pa_ess/http_updater.ex
+++ b/lib/pa_ess/http_updater.ex
@@ -15,7 +15,7 @@ defmodule PaEss.HttpUpdater do
           {:ok, :sent} | {:ok, :no_audio} | {:error, :bad_status} | {:error, :post_error}
 
   # Normally an anti-pattern! But we mix compile --force with every deploy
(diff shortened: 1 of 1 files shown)


PaulJKim:
Assigned to @cmaddox5 since this application is owned by the screens team now, but still added @losvedir as a reviewer since the hand-off is still fresh.

PaulJKim:
On second thought, there's no need to merge this since I can just deploy the branch and have it run against the ARINC test server. Closing out the PR

PR #474, opened 2022-01-18
Add new glx stations

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Updates to realtime-signs](https://app.asana.com/0/1199504266721641/1200959142579185/f)

This PR adds source configs for the new Lechmere and Union Square stations. Also adds a new headsign name in `PaEss.Utilities` for Union Sq, which will only be used for Lechmere until the new stations open for revenue service.

#### Reviewer Checklist

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6239,12 +6239,22 @@
     "pa_ess_loc": "GUNS",
     "read_loop_offset": 30,
     "text_zone": "e",
-    "audio_zones": ["e"],
+    "audio_zones": [],
     "type": "realtime",
     "source_config": [
       [
         {
           "stop_id": "70503",
+          "direction_id": 0,
+          "headway_direction_name": "Heath Street",
+          "routes": ["Green-E"],
+          "platform": null,
+          "terminal": true,
+          "announce_arriving": false,
+          "announce_boarding": false
+        },
+        {
+          "stop_id": "70504",
           "direction_id": 0,
           "headway_direction_name": "Heath Street",
           "routes": ["Green-E"],
@@ -6260,12 +6270,22 @@
     "id": "union_sq_west",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
-    "read_loop_offset": 90,
+    "read_loop_offset": 30,
     "text_zone": "w",
-    "audio_zones": ["w"],
+    "audio_zones": ["e", "w", "m"],
     "type": "realtime",
     "source_config": [
       [
+        {
+          "stop_id": "70503",
+          "direction_id": 0,
+          "headway_direction_name": "Heath Street",
+          "routes": ["Green-E"],
+          "platform": null,
+          "terminal": true,
+          "announce_arriving": false,
+          "announce_boarding": false
+        },
         {
           "stop_id": "70504",
           "direction_id": 0,
@@ -6283,9 +6303,9 @@
     "id": "union_sq_mezzanine",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
-    "read_loop_offset": 120,
+    "read_loop_offset": 30,
     "text_zone": "m",
-    "audio_zones": ["m"],
+    "audio_zones": [],
     "type": "realtime",
     "source_config": [
       [
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6234,7 +6234,7 @@
     ]
   },
   {
-    "id": "union_sq_north",
+    "id": "union_sq_track_one",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
     "read_loop_offset": 30,
@@ -6267,7 +6267,7 @@
     ]
   },
   {
-    "id": "union_sq_south",
+    "id": "union_sq_track_two",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
     "read_loop_offset": 90,
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6234,7 +6234,7 @@
     ]
   },
   {
-    "id": "union_sq_east",
+    "id": "union_sq_north",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
     "read_loop_offset": 30,
@@ -6267,7 +6267,7 @@
     ]
   },
   {
-    "id": "union_sq_west",
+    "id": "union_sq_south",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GUNS",
     "read_loop_offset": 30,
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6155,7 +6155,30 @@
     ]
   },
   {
-    "id": "lechmere_green_line",
+    "id": "lechmere_green_line_eastbound",
+    "headway_group": "green_trunk",
+    "pa_ess_loc": "GLEC",
+    "read_loop_offset": 30,
+    "text_zone": "e",
+    "audio_zones": ["e"],
+    "type": "realtime",
+    "source_config": [
+      [
+        {
+          "stop_id": "70501",
+          "direction_id": 1,
+          "headway_direction_name": "Eastbound",
+          "routes": ["Green-D", "Green-E"],
+          "platform": null,
+          "terminal": false,
+          "announce_arriving": false,
+          "announce_boarding": false
+        }
+      ]
+    ]
+  },
+  {
+    "id": "lechmere_green_line_westbound",
     "headway_group": "green_trunk",
     "pa_ess_loc": "GLEC",
     "read_loop_offset": 90,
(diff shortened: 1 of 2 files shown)


priv/signs.json
```diff
@@ -6155,17 +6155,152 @@
...
+          "platform": null,
+          "terminal": false,
+          "announce_arriving": false,
+          "announce_boarding": false
+        }
+      ]
+    ]
+  },
+  {
+    "id": "union_sq_east",
```
losvedir, line 6237: You know, come to think of it, it seems kind of odd to me that we have separate east, west, and mezzanine zones at Union Sq since I'd expect all signs to display the same data: trains towards Heath St. We don't generally show arrival predictions for when trains will get to the terminal.

We've asked Andrew to consolidate zones before (and break apart zones), so that might be worth considering here.

As long as they're separate zones, you might want to do something like: give them all the same `source_config`, and give one of the zones `"audio_zones": ["e", "w", "m"]`, and the others `[]`. That way, it won't be reading the same information all the time, and when it does read it out, it will send the message like `sta=GUNS111000` as a single request, which should make sure they play in sync.

losvedir: As for what that `source_config` should be, that's still an open question. We haven't decided yet on the realtime team exactly what stop IDs we'll use. Static GTFS uses `70503` for arrivals and `70504` for departures, but operationally there are two platforms there and either platform could handle the train pulling in (and then subsequent departure).

So you might want a `source_config` of two entries, `70503` and `70504` both in the `0` direction, and so it would combine the predictions out of either platform.

However, we're also still considering an approach like we do for Alewife, Forest Hills, and Oak Grove, where trains are predicted _to_ `70503` and reverse predictions would be _from_ `70504`, but then when the train is actually _at_ the platform there's be yet two more stop IDs `Union Square-01` and `Union Square-02`. And in that case, you'd want all four stop IDs in your `source_config` list!

I think for now it's good to just go with the first approach (`70503` and `70504` in the `0` direction), and if we go with `Union Square-01` and `Union Square-02`, we'll let you know and you can add them as well.

PaulJKim: Thank you for the clarification and suggestions! Super helpful especially since I definitely am still getting a handle on how exactly the configs here will play out in real life at the stations.

Re: the audio zones, do you think it might be worth it to have another sign zone have all of the audio zones as well? Just thinking that if only one of the signs has all three audio zones, the audio might play pretty infrequently. Let me know your thoughts there!

losvedir: The read loop triggers the sign to read once every 4 minutes, reset by any "interrupting reads" from things like the sign going to "now arriving" or "stopped n stops away". That's the value used by all signs throughout the system. If the signs at Union Square are all configured to show the same information (which I'm guessing they should, since they [all look close to each other](https://upload.wikimedia.org/wikipedia/commons/thumb/a/a4/Union_Square_station_construction_%283%29%2C_December_2021.JPG/1200px-Union_Square_station_construction_%283%29%2C_December_2021.JPG) and the trains only go one way), then conceptually they're just a single zone and should read at once / 4 minutes.

Adding audio zones to another sign will either have them start out on a sort of heartbeat cadence, depending on the `read_loop_offset`, as they'd each read every 4 minutes staggered by their offset. But then an "interrupting read" would happen and they'd all be on the same every-4-minute cycle anyway.

PaulJKim: Makes sense!

Re: The source config, we'd still want the `text_zone` and `id` to be different from each other, since that's what allows us to specify which signs we're talking to right? At least until we decide if we want to ask Andrew to combine the zones into one and regardless of the fact that they're all showing the same information for stop_id 70503 and 70504

In other words, for now their source configs should be identical except we'd still want separate `id`s (such as `union_sq_west`) and their respective `text_zone` values?

losvedir: Yes, sorry by `source_config` I meant specifically just the data in the `"source_config"` key. I think we call the overall thing the "sign config" maybe? I admit it's a bit confusing. I think the signs should have identical `"source_config"` lists, but yes, the `"id"`, `"text_zone"`, and `"audio_zones"` would be different.

PaulJKim: Going back to the topic of the announcements at the new Union Sq station. I know that at other terminals, we keep the signs in headway mode although I'm unsure if that's only where there is a train yard present. Would that be the case at Union Sq, or would we want to announce departures or something since there's no train yard (I think) at Union Sq and I'm guessing that in that case, the trains just turn around and go the other way?

losvedir: We've currently set expectations as the signs there only being in headway mode. There is a yard at Union Square station (though inbound and not *past* the station). So there's a potential for trains to arrive at Union Sq, offload passengers, and then start heading back empty and then pull into the yard.

Furthermore, we don't really know operationally what will be going on there. After a train arrives and offloads passengers, it will probably wait some amount of time, but we don't know if that will be a consistent amount of time or very accurately predictable. And sadly we don't have "train sheets" like we do for the heavy rail where the dispatchers tee up the departures and send us messages "delay the trip departure by 3 minutes" or "planning to leave one minute early". There is the Glides program here in CTD that's working on, among other things, just that, though! So hopefully someday we'll have good departure predictions.

PaulJKim: I see, so sounds like keeping the zones as configured to not provide arrival/departure messages is correct given the current expectations and state of the system 😄

priv/signs.json
```diff
@@ -6165,7 +6188,139 @@
...
+      ]
+    ]
+  },
+  {
+    "id": "union_sq_west",
+    "headway_group": "green_trunk",
+    "pa_ess_loc": "GUNS",
+    "read_loop_offset": 30,
+    "text_zone": "w",
+    "audio_zones": ["e", "w", "m"],
```
digitalcora, line 6275: Any particular reason to attach the audio zones to the "west" config? Not that there's any functional difference, but I might have expected it to be either the first Union Square config in the file (east), or the "middle" one (mezzanine).

priv/signs.json
```diff
@@ -6165,7 +6188,139 @@
...
+          "announce_boarding": false
+        }
+      ]
+    ]
+  },
+  {
+    "id": "union_sq_west",
+    "headway_group": "green_trunk",
+    "pa_ess_loc": "GUNS",
+    "read_loop_offset": 30,
```
digitalcora, line 6273: Should we also give these the "standard" read loop offsets (seems to be 90 for west, 120 for mezzanine)? Even if nothing will actually be read for now, it could avert future issues if we split up the audio zones later.

priv/signs.json
PaulJKim: Mainly just because all of the predictions that we'd be showing for this station would be "westbound", but that's a pretty loose reasoning. But it does make a little bit more sense to me to put it with the mezzanine zone now that I think about it

priv/signs.json
PaulJKim: Would the signs not be reading out the headways or anything else at all? I figured that it keeping them on the same loop makes sense to avoid over frequently reading out the same content since they're all showing the same thing, but definitely correct me if I'm misinterpreting how the read loops would be behaving here

digitalcora: The way you currently have it set up, all the audio zones are associated with just one of the configs. The other configs aren't connected to any audio zones, so their "read loop" is irrelevant — they _have_ an offset, because the format of the config requires it, but can't play audio since they aren't connected to any audio zones. At least, that's my understanding of how the audio system works, and my interpretation of @losvedir's comments above, but I should caution that I'm not an expert on this.

PaulJKim: Oh yes, that makes total sense. I was just confusing myself. Since they don't have any audio zones set up, then there wouldn't be any audio playing for those zones :)

priv/signs.json
```diff
@@ -6165,7 +6188,139 @@
...
+          "platform": null,
+          "terminal": false,
+          "announce_arriving": true,
+          "announce_boarding": false
+        }
+      ]
+    ]
+  },
+  {
+    "id": "union_sq_north",
```
digitalcora, line 6237: Should be `union_sq_track_one` going by the signs_ui changes (and likewise for south => track_two).

PaulJKim: Just pushed that change

digitalcora (approved):
Seems ok to me to start testing in dev! Of course, if you can wait until tomorrow morning, @losvedir could also give it a final once-over.

losvedir (approved):
I think this looks good and should be safe to deploy. When the branch opens you can update the other E-branch signs with the new headway destination of Union Square.

Now that there's a new destination of `"Union Square"` there are a number of messages to update. For example check out `TrainIsArriving.audio_params/1` or `VehiclesToDestination.message_id`. Sadly, there's not a centralized listing of them or what exactly the message IDs are! But if you poke around in `lib/content/audio/` you should find all the places they need to be updated. :-/

Another thing you can do is update a few signs like Kenmore to have a headway destination of Union Square on dev and let it run for a while. If Realtime Signs fails to look up a message ID that it needs, it will log a warning about it. So you can let it run for a while and then see what gets logged.

I forget what message ID resources you have. If you can't find the various message IDs or what exactly the audio they correspond to is, let me know and I can dig around. Or maybe best for you to email Andrew, asking for an official and up to date export which you can then put somewhere for your team, and maintain.

PR #475, opened 2022-01-20
Update Realtime Signs to log UTC time

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Change Realtime Signs to log in UTC](https://app.asana.com/0/0/1201681770685699/f)

This change is so that the logger for Realtime Signs is configured to log UTC time rather than local time.

Before:
Screen Shot 2022-01-20 at 6 05 37 PM

After:
Screen Shot 2022-01-20 at 6 06 06 PM

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -49,7 +49,7 @@ config :ex_aws,
   access_key_id: [{:system, "SIGNS_S3_CONFIG_KEY"}, :instance_role],
   secret_access_key: [{:system, "SIGNS_S3_CONFIG_SECRET"}, :instance_role]
 
-config :logger, backends: [:console]
+config :logger, backends: [:console], utc_log: true
 
 config :ehmon, :report_mf, {:ehmon, :info_report}
 


losvedir (commented):
This addresses part of the issue, but only the `:console` logger, which we log at the `:warn` level to disk.

You should also update the `logger_splunk_backend` logger (check `config/prod.exs`).

The issue we have faced specifically is during DST "fall back" time. Something like "1:05am" is ambiguous (is it the first or the second 1:05am?). And what happens during that time is by some mechanism I haven't traced out, any logs that Realtime Signs does during the first 1-2am are sent "forward in time". That is in Splunk, during the first 1am hour, you don't see any logs coming in. But actually if you expand the time frame to include the future, you'll see that the logs *are* appearing, just one hour in the future.

So what seems to be happening is that somewhere along the line, we're trying to log "1:05am" local time, it gets converted to the wrong one, and that gets sent to Splunk. But if the app, or the logging, or the dep is in UTC time, then that won't be an issue. I'm not exactly sure where the final fix belongs or how it looks, though. One idea is setting the `TZ` environment variable in the WinSW config to UTC time (off the top of my head I think something like `Etc/UTC`, which will cause Erlang/Elixir in the app to operate in UTC time.

PaulJKim:
Gotcha. Thanks for the heads up about the splunk backend. And good point about the latency script. Will probably circle back to this post GLX testing

cmaddox5 (approved):

PR #476, opened 2022-01-31
Revert "fix: Alert display logic given Brookline Hills closure"

PaulJKim:
Reverts mbta/realtime_signs#451

We removed Brookline Hills when it was being rebuilt and it was removed from GTFS and Alerts, but we need to add it back. Bug was reported by Eryk Rotondo in [slack thread](https://mbta.slack.com/archives/GKQ8R5HAN/p1643458793570459) letting us know that brookline hills is set to auto but was blank instead of showing the appropriate alert message.


digitalcora (approved):
Good sleuthing! :shipit:

PR #477, opened 2022-02-10
Add uses shuttle flag

PaulJKim:
#### Summary of changes

An ask was brought up by PIOs (Eryk Rotondo) to set a new default message at Beaconsfield on the D Line when the alert is a shuttle alert that omits mention of shuttles. This is because the roads around that stop are too narrow and shuttles never run there.

This PR adds an additional flag `uses_shuttles` and sets the values for the Beaconsfield to `false`. The logic was adjusted to account for this flag and return a different message when the flag is false.

Screenshot of the change running in **Dev**
Screen Shot 2022-02-10 at 3 02 27 PM

#### Reviewer Checklist

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -50,28 +50,39 @@ defmodule Signs.Utilities.Messages do
           Engine.Alerts.Fetcher.stop_status()
         ) :: sign_messages()
   defp get_headway_or_alert_messages(sign, current_time, alert_status) do
-    case get_alert_messages(alert_status) do
+    case get_alert_messages({alert_status, sign.uses_shuttles}) do
       nil -> Signs.Utilities.Headways.get_messages(sign, current_time)
       messages -> messages
     end
   end
 
-  @spec get_alert_messages(Engine.Alerts.Fetcher.stop_status()) :: sign_messages() | nil
-  defp get_alert_messages(alert_status) do
-    case alert_status do
-      :shuttles_transfer_station ->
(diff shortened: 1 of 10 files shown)


jzimbel-mbta (approved):
LGTM! Might want to think about refactoring if we end up having another "flag" argument like this that only serves certain rare edge cases, as they can quickly gunk up the readability of the code, but with just one this change is fine.

lib/signs/utilities/messages.ex
```diff
@@ -50,28 +50,32 @@ defmodule Signs.Utilities.Messages do
...
 
-  @spec get_alert_messages(Engine.Alerts.Fetcher.stop_status()) :: sign_messages() | nil
-  defp get_alert_messages(alert_status) do
-    case alert_status do
-      :shuttles_transfer_station ->
+  @spec get_alert_messages(Engine.Alerts.Fetcher.stop_status(), boolean()) ::
+          sign_messages() | nil
+  defp get_alert_messages(alert_status, uses_shuttles) do
+    case {alert_status, uses_shuttles} do
+      {:shuttles_transfer_station, _} ->
```
hannahpurcell, line 63: `:shuttles_transfer_station` is a new case you've added here.  What does that mean?  How is it different from `:suspension_transfer_station`?

PaulJKim: that one already existed! I think it just looks like an entirely new case with how Git is diffing the versions. But to clarify the different `alert_status`es, `:shuttle_transfer_station` is basically when there's no train service running but there are shuttles for riders to transfer to and `:suspension_transfer_station` would mean there is no service but also no shuttles. I'm not exactly sure in what real world situations the suspension alert is used though.

So the

PR #481, opened 2022-03-07
Update headways for glx revenue service

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Update e-line with new headway directions to Union Square](https://app.asana.com/0/1185117109217413/1201714031001944/f)

This PR adds updates to the config and certain audio-related modules which will enable Union Square as a destination and headway direction on the E-Line for PA/ESS. This change shouldn't be deployed until revenue service is officially beginning, which is currently planned for March 21st, 2022.

### DIFF
--- a/lib/content/audio/train_is_boarding.ex
+++ b/lib/content/audio/train_is_boarding.ex
@@ -47,7 +47,14 @@ defmodule Content.Audio.TrainIsBoarding do
     end
 
     defp do_to_params(%{destination: destination, route_id: "Green-" <> _branch}, destination_var)
-         when destination in [:lechmere, :north_station, :government_center, :park_st, :kenmore] do
+         when destination in [
+                :lechmere,
+                :north_station,
+                :government_center,
+                :park_st,
+                :kenmore,
+                :union_square
+              ] do
       vars = [
         @the_next,
         @train_to,
--- a/lib/content/utilities.ex
+++ b/lib/content/utilities.ex
@@ -69,7 +69,7 @@ defmodule Content.Utilities do
   def destination_for_prediction("Green-B", 1, _), do: {:ok, :government_center}
   def destination_for_prediction("Green-C", 1, _), do: {:ok, :government_center}
   def destination_for_prediction("Green-D", 1, _), do: {:ok, :north_station}
-  def destination_for_prediction("Green-E", 1, _), do: {:ok, :north_station}
+  def destination_for_prediction("Green-E", 1, _), do: {:ok, :union_square}
 
   def destination_for_prediction(_, _, _), do: {:error, :not_found}
 
--- a/lib/content/utilities.ex
+++ b/lib/content/utilities.ex
@@ -54,6 +54,7 @@ defmodule Content.Utilities do
 
   def destination_for_prediction(_, 1, "70209"), do: {:ok, :lechmere}
   def destination_for_prediction(_, 1, "70205"), do: {:ok, :north_station}
+  def destination_for_prediction(_, 1, "70503"), do: {:ok, :union_square}
   def destination_for_prediction(_, 1, "70201"), do: {:ok, :government_center}
   def destination_for_prediction(_, 1, "70200"), do: {:ok, :park_street}
   def destination_for_prediction(_, 1, "71199"), do: {:ok, :park_street}
--- a/lib/content/utilities.ex
+++ b/lib/content/utilities.ex
@@ -52,7 +52,6 @@ defmodule Content.Utilities do
   def destination_for_prediction(_, 0, "70161"), do: {:ok, :riverside}
   def destination_for_prediction(_, 0, "70260"), do: {:ok, :heath_street}
 
-  def destination_for_prediction(_, 1, "70209"), do: {:ok, :lechmere}
   def destination_for_prediction(_, 1, "70205"), do: {:ok, :north_station}
   def destination_for_prediction(_, 1, "70503"), do: {:ok, :union_square}
   def destination_for_prediction(_, 1, "70201"), do: {:ok, :government_center}
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6134,7 +6134,7 @@
         {
           "stop_id": "70207",
           "direction_id": 1,
-          "headway_direction_name": "Eastbound",
+          "headway_direction_name": "Union Square",
           "routes": ["Green-B", "Green-C", "Green-D", "Green-E"],
           "platform": null,
           "terminal": false,
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -1024,7 +1024,7 @@
           "announce_boarding": true,
           "source_for_headway": true
         },
-	{
+        {
           "stop_id": "Oak Grove-01",
           "routes": ["Orange"],
           "direction_id": 0,
@@ -1033,8 +1033,8 @@
           "terminal": true,
           "announce_arriving": false,
           "announce_boarding": true
-	},
-	{
+        },
+        {
           "stop_id": "Oak Grove-02",
           "routes": ["Orange"],
           "direction_id": 0,
@@ -1068,7 +1068,7 @@
           "announce_boarding": true,
           "source_for_headway": true
         },
-	{
+        {
           "stop_id": "Oak Grove-01",
           "routes": ["Orange"],
           "direction_id": 0,
@@ -1077,8 +1077,8 @@
           "terminal": true,
           "announce_arriving": false,
           "announce_boarding": true
-	},
-	{
+        },
+        {
           "stop_id": "Oak Grove-02",
           "routes": ["Orange"],
           "direction_id": 0,
@@ -1112,7 +1112,7 @@
           "announce_boarding": true,
           "source_for_headway": true
         },
-	{
+        {
           "stop_id": "Oak Grove-01",
           "routes": ["Orange"],
           "direction_id": 0,
@@ -1121,8 +1121,8 @@
           "terminal": true,
           "announce_arriving": false,
           "announce_boarding": true
-  },
-	{
+        },
+        {
           "stop_id": "Oak Grove-02",
           "routes": ["Orange"],
           "direction_id": 0,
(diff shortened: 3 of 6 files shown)


priv/signs.json
```diff
@@ -6019,7 +6019,7 @@
         {
           "stop_id": "70205",
           "direction_id": 1,
-          "headway_direction_name": "Lechmere",
+          "headway_direction_name": "Union Square",
           "routes": ["Green-B", "Green-C", "Green-D", "Green-E"],
```
PaulJKim, line 6023: @losvedir if I could ask you a question here: Should we keep `"Green-B", "Green-C", "Green-D"` as part of the `routes` list since North Station is still part of green trunk? Or should we remove those for this sign since only E-line trains are going to be heading eastbound from here. Not sure if we would want alerts affecting this sign for branches that terminate at North Station. Although, I noticed Green-B is in here while the B Line doesn't go to North Station, so maybe we just include them all for any green trunk stop?

losvedir: The `routes` list serves two purposes.

The first is with alerts, as you've alluded to. But I don't think a `Green-B` alert would affect the sign out here, unless its stop ID was included for some reason in the alert (if that's even possible). I think it _might_ show up if it was an entire _route_ suspension, which actually might be bad because it would say "No train service", even though technically there could still be Union Sq <-> Park St service in that case.

However, that field is also [used for filtering predictions](https://github.com/mbta/realtime_signs/pull/166). Back when I did that PR, Kenmore station had two different berths that trains stopped at (one for Green-B and one for Green-C/Green-D) , with corresponding countdown signs for each, but only one underlying GTFS stop ID. So to filter the predictions that appeared on the sign we used a combination of the stop ID and the route, to get the right train predictions to the right sign. However, since then we actually created different stop IDs for the different berths, and you can see here in `signs.json` that one sign uses `71150` and the other `70150`. So we could probably omit the filtering functionality now. I can't think of any other instances on the subway where we have 1) different countdown signs requiring different boarding information and 2) the same underlying stop ID. Stations like Downtown Crossing with both Red and Orange service have different stop IDs, as does Park St with its different berths in both the eastbound and westbound directions.

But unless or until you remove the filtering functionality, you probably want to include all the routes here. It's _possible_ that occasionally you'll have Green-B/C/D predictions out here. It's more likely inbound, as a train could be told to go to Boston College / Cleveland Circle / Riverside, and the operator will key in their new AVI code, and we'll start predicting them to go out that way. An outbound train could possibly have a B/C/D route keyed in either through operator error or because they're planning to go out to Union Sq and then back to B/C/D branches.

losvedir: For instance, over the last two weeks, you can see we've [made a handful of predictions](https://prediction-analyzer.mbtace.com/accuracy?filters[bin]=All&filters[chart_range]=Daily&filters[date_end]=2022-03-16&filters[date_start]=2022-03-02&filters[direction_id]=any&filters[in_next_two]=&filters[mode]=subway&filters[route_ids]=Green-B&filters[service_date]=2022-03-15&filters[stop_ids][]=70204) for Green-B trains at Haymarket.

lib/content/utilities.ex
```diff
@@ -54,6 +54,7 @@ defmodule Content.Utilities do
 
   def destination_for_prediction(_, 1, "70209"), do: {:ok, :lechmere}
   def destination_for_prediction(_, 1, "70205"), do: {:ok, :north_station}
+  def destination_for_prediction(_, 1, "70503"), do: {:ok, :union_square}
   def destination_for_prediction(_, 1, "70201"), do: {:ok, :government_center}
```
hannahpurcell, line 58: The spec for this function isn't explicit.  Parameters are:  route_id, digit to represent directionality of the terminus, stop_id?

hannahpurcell (commented):
Couple questions for starters.  Otherwise, I'm ready to approve, pending the question to Gabe.

lib/content/utilities.ex
```diff
@@ -54,6 +54,7 @@ defmodule Content.Utilities do
 
   def destination_for_prediction(_, 1, "70209"), do: {:ok, :lechmere}
```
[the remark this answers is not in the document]
losvedir: Ah, good catch! `70503` is one of the _old_ Lechmere stop IDs. With GLX the station was rebuilt nearby and so got new stop IDs (the ones in the API).

So here, the options are: Update the stop ID to the new one, or remove this line entirely. The difference is how signs should behave if for some reason we make predictions for a trip where those predictions end at Lechmere. One reason why that might happen is if there's bus shuttling between Lechmere and Union Sq. If we list the new Lechmere stop ID then the signs will show "Lechmere 5 min" (or whatever), but if we keep it omitted the signs  will say "Union Sq 5 min".

We handle it both ways in other cases: An eastbound (from the branches) Green-E train terminating for some reason at Park St rather than North Station, will show up on signs as "Park St  X min", since Park St is a known station and a common terminus for other lines. However, a westbound (from the trunk) Green-E train terminating at, say, Brigham Circle, will show up on signs as "Heath St   X min". We do that when the stations are notable enough, with the idea that if passengers are trying to get to Heath St, they might be looking for a "Heath St" train (and then will face an unexpected early stop and shuttle, but will still get there) vs not knowing that Brigham Circle is even on the E branch.

My personal opinion is that now that Lechmere is no longer a terminal with a yard, we don't expect trains to ever terminate at Lechmere during normal service, and so it's more in the latter camp, and should show up on signs as "Union Sq", implying that this line should simply be deleted.

lib/content/utilities.ex
PaulJKim: Yep basically. I guess it'd be more accurate to say `Line name`, or whatever the more technical term is. I _think_ a route Id isn't necessarily synonymous with the name of the line, but I may be wrong. Yep, the bit represents the directionality, and the stop_id is used for specific cases where a stop is a commonly used terminus. If the predictions are not for a terminus defined in these specific cases, it'll default to using the terminus for the line which are defined in lines 66-73 for the Green line, for example

lib/content/utilities.ex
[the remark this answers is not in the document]
PaulJKim: Agreed, I think it's probably best to remove Lechmere as a terminal. Thank you for catching this @hannahpurcell !

lib/content/utilities.ex
losvedir: The first parameter is actually the route ID here. "Line" is an "experimental" GTFS extension that [the MBTA uses](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#linestxt), and would be [`line-Green`](https://api-v3.mbta.com/lines/line-Green), referring to the set of routes, `Green-B`, `Green-C`, `Green-D`, and `Green-E`.

This function actually does match against route as the first argument. (You'll see some clauses that match against `Green-B`, for example. Others match against `Red`, but confusingly `Red` is a route (with `line-Red` as the line ID).

In this set of functions we omit the route altogether so that, e.g., `"70205"` for _any_ route will match, but really we mean any Green line route (and no other route will have that stop ID). For clarity, the pattern could be `"Green-" <> _`, I suppose.

priv/signs.json
PaulJKim: Updating on this: I had a separate PR that removed the other routes from the config for Science Park, but I think it's best to just leave them since I think it'll affect the station in the same way as mentioned by Gabe above. Meaning - alerts from other lines probably won't affect Science Park unless for some reason the stop_id was included in the alert and in the slight chance that an inbound train at Science Park gets told to go to a different branch, those predictions would show on the Science Park signs.

hannahpurcell (approved):
I'm ready to approve on my end 👍

priv/stops.json
```diff
@@ -9,10 +9,6 @@
     {"station": "G Government Center eastbound", "stop_ids": ["70201"]},
     {"station": "G Haymarket eastbound", "stop_ids": ["70203"]},
     {"station": "G North Station eastbound", "stop_ids": ["70205"]},
-    {"station": "G Science Park eastbound", "stop_ids": ["70207"]},
```
PaulJKim, line 12: Apologies for missing this, but while checking my work on this PR, I realized that I missed adding the new stops to these lists here. If I'm not mistaken, I think these are used for alerts so definitely important that they go out on Monday @losvedir @hannahpurcell

lib/content/audio/train_is_boarding.ex
```diff
@@ -47,7 +47,14 @@ defmodule Content.Audio.TrainIsBoarding do
     end
 
     defp do_to_params(%{destination: destination, route_id: "Green-" <> _branch}, destination_var)
-         when destination in [:lechmere, :north_station, :government_center, :park_st, :kenmore] do
+         when destination in [
```
PaulJKim, line 50: Also realized I missed adding `:union_square` here. I believe the boarding messages for Union Sq would have been handled by the ad hoc logic anyway, but we have actual recordings for Union Square so we should use them. @losvedir @hannahpurcell

priv/stops.json
losvedir: Oh, yep, good catch, this file should be updated. This file is used to calculate which stations are adjacent to other stations, for the purposes of what to display on an alert. Namely, if a given station has a shuttle alert, along with all its neighbors, then we say that the station is closed. If one of its neighbors has service, then we consider the station an "endpoint" of the shuttle stretch, and we just blank the sign, expecting the PIO to put something useful on it about where or how to catch the shuttles.

(Alas, this would be better solved by an inference of an alert's activities, which specifies whether a given stop ID allows boarding, exiting, or riding, but that either didn't exist yet or didn't occur to us at the time we wrote this.)

In that regard, the stations being adjacent to each other in the list is important, and in fact we only use the list grouping names ("Green-E-eastbound") for organization; they don't really matter.

So I think the fix here would be to not delete these rows (even though technically the stations are not in the Green-Trunk), but to update Lechmere's stop IDs, and then add in Union Sq as two rows, between the two Lechmeres. (The list is kind of an outbound-then-inbound loop.)

PaulJKim: Gotcha, that makes sense. I was just checking out the unit test that's failing after adding them to the new stops and agreed that the fix is to change the IDs then add Union Sq in the middle.

PR #483, opened 2022-03-15
Update headway groups for GLX and science park config

PaulJKim:
#### Summary of changes
[**Asana Ticket:** [Update e-line with new headway directions to Union Square](https://app.asana.com/0/1185117109217413/1201714031001944/f)]

This PR includes config changes to move the GLX stations (Including Science Park) from the Green Trunk headway group to the Green E headway group. This makes sense because, at least until the second GLX branch opens, only E-Line trains will run through Science Park and on GLX.

Once D-Line trains begin running for revenue service, we may want to reconsider what headway groups to include Science Park and Lechmere in.

If wondering about what all of the different config values do, `lib/signs/utilities/source_config.ex` contains details on each of them


jzimbel-mbta (approved):
Looks good!

Side note, I love that Union Square's ID is "GUNS" 😂

PR #484, opened 2022-03-18
Make union sq zones to be single source

PaulJKim:
#### Summary of changes
Initially these sign zones were set up with having two sources for predictions. For one, we aren't giving predictions on GLX yet, and having two sources configured causes the headway messages for a sign to omit the headway direction. For example, the sign would show "Trains every 5 to 7 min" instead of "Heath St trains every 5 to 7 min.

This change updates the Union Sq zones to be single sourced so that they include "Heath St" for headway messages. We should be sure to loop back around to this and reconfigure the signs once we understand how predictions will be provided for this stop.

Asana Ticket filed for follow-up work once predictions are available: https://app.asana.com/0/1201753694073608/1201979909970000/f

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -1024,7 +1024,7 @@
           "announce_boarding": true,
           "source_for_headway": true
         },
-	{
+        {
           "stop_id": "Oak Grove-01",
           "routes": ["Orange"],
           "direction_id": 0,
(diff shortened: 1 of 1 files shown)


jzimbel-mbta (approved):
This looks good! If you haven't already, could you file an asana task for the followup work to make sure we don't forget to do it, and maybe also link it in this PR's description?

PaulJKim:
good call! I will do that now before merging

PR #487, opened 2022-03-25
Improve UIDs for Realtime Signs

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Realtime Signs UID improvement](https://app.asana.com/0/1201753694073608/1201992374136791/f)

**Background**
Currently in Prod, we run a pool of 16 `PaEss.HttpUpdater` GenServers that parallelize our sending of visual and audio message requests to ARINC's headend server. Each of these updaters maintains its own counter for UIDs which means that currently, there can be up to 16 duplicates for a given UID that we use in our requests to ARINC in prod.

Now that we're working with ARINC to improve logging/monitoring of PA/ESS, in order to enable ARINC to easily correlate and compile logs between the Headend server and all of the SCUs, we should send them a sufficiently Unique ID that is traceable at least within a days worth of logs. The reason why the timeframe is one day is that ARINC pulls all the SCU logs into the headend server every night around 1:30AM. At that point, we intend to pull their compiled logs and dump them into Splunk.

**Details**
The GenServer being added is a single, simple counter that all of the HttpUpdaters will request to get a UID to use in their requests to ARINC. To attain uniqueness across deploys, we have a deploy counter which gets read from a text file, incremented, and re-written to the text file when the GenServer is started. In the counter, we append the deploy number to each Id so that on the first deploy, the UIDs would be as follows:
`100`, `200`, `300`...
And on the subsequent deploy the UIDs would be:
`101`, `201`, `301`... and so forth.

Once the counter reaches 99, it would be reset to zero. 100 deploys gives us plenty of leeway for making sure that deploys stay unique within a 24 hour period.


paulswartz:
Are there limits to how large the UID can be, and does it have to be an integer? There are some existing libraries which can generate unique IDs and don't require maintaining local state:
- https://hex.pm/packages/snowflake
- https://hex.pm/packages/nanoid
- https://hex.pm/packages/cuid

PaulJKim:
@paulswartz Andrew Daniels from ARINC did say that the UID needs to be an int. I suppose that implies that the ID can't be larger than 2^32

PaulJKim:
@paulswartz No problem! And yes, so we have a maximum limit, which we define [here](https://github.com/mbta/realtime_signs/blob/main/lib/pa_ess/http_updater.ex#L18), of 32 messages per second. Doing some spot checks with Splunk though, it seems like we rarely approach that limit and hover somewhere between 10-20 messages per second even during peak hours.

paulswartz:
Could you do something like:
```elixir
> timestamp = div(System.system_time(:millisecond), 500)
> updater_index = 5 # 0..15
> sequence = 4 # 0..31
> << uid::unsigned-integer-31 >> = << timestamp::unsigned-integer-22, updater_index::unsigned-integer-4, sequence::unsigned-integer-5>>
> uid
140048548
```

This gives you a 31-bit integer (no worries about signed/unsigned), with 16 unique updaters and and 32 messages per timestamp, without needing to maintain the deploy index. 22 bits of half-second granularity gives you a bit over 3 weeks until the UIDs would repeat. If you're sure ARINC supports 32-bit integers, you'd get almost 7 weeks before a repeat.

Alternatively, if you have one process generating the IDs (as you do here), you could drop the `updater_index` and have over a year before the UIDs repeat.

PaulJKim:
@paulswartz Just making sure I'm following the math here: so div(System.system_time(:millisecond), 500) gives us a 32 bit integer. Would your solution be to grab 22 bits of that, and then use 4 bits to indicate which updater, and then use the last 5 bits as a counter to 32? And then either each updater would have its own counter to 32 and get a new timestamp once the counter resets, or we could have a single process like we do now and omit the counter?

As a side note, we are actually using 24 updaters in production so I think we would need 5 bits for that. The UIDs only need to be unique over the course of 1 day at the minimum though, although I think it doesn't hurt to have them be unique for longer, so I think we could steal a bit from somewhere no problem.

Am I understanding that correctly?

PaulJKim:
@paulswartz Oh right, I had the counter and updater index backwards in my head. All makes sense though, and I agree that avoiding the use of state is preferable for this. 

And just to clarify, the internal counter (whether the logic is in individual updaters or in a single process) is basically to guarantee that enough time has passed by the next time we grab a new timestamp, especially since we're moving the granularity up to half-seconds?

paulswartz:
Correct: even if the granularity was microseconds, you could have multiple requests in the same microsecond, and the counter ensures that this doesn't generate a non-unique ID. 

The approach that the `snowflake` library uses is to keep track of the last timestamp as well as the counter, and when an ID is requested, reset the counter to `0` if the timestamp has changed. This results (in practice) with the counter only being incremented when there are multiple UIDs generated within the same timestamp window.

PaulJKim:
That's a neat approach. Thank you for the input on this @paulswartz! I will rework this so we don't have to use a text file to track the deploy count.

losvedir:
This is a clever approach! We had talked about taking the timestamp to handle the deploy issue, but it used up too many bits, and I didn't think of just taking the lower bits.

Question: why are we going to 1/2 second resolution? Why not second resolution, given that we have the `sequence` room?

Also, I have a vague feeling that some amount of time resolution doesn't work right on Windows - that might be something to check. @PaulJKim - can you pop open a Git Bash shell and then `iex` on opstech3, and verify that `System.system_time(:millisecond)` works as we're expecting here?

PaulJKim:
@losvedir It does seem to work on iex on git bash on opstech3. Not sure I follow what you mean by time resolution though?

paulswartz:
> Question: why are we going to 1/2 second resolution? Why not second resolution, given that we have the `sequence` room?

I went with 1/2 second resolution to give some wiggle room with the 32 messages/second count. If you go to full second resolution, I'd consider giving an extra bit to the counter.
 
> Also, I have a vague feeling that some amount of time resolution doesn't work right on Windows - that might be something to check. @PaulJKim - can you pop open a Git Bash shell and then `iex` on opstech3, and verify that `System.system_time(:millisecond)` works as we're expecting here?

IIRC it was something like 16ms resolution? So it's not exactly millisecond resolution, but close enough for ~government~ transit work.

PaulJKim:
@paulswartz @losvedir I've updated this PR so that each updater generates its own uids with the use of a updater index and internal counter in conjunction with a timestamp in half-seconds. Would appreciate another pass from either of you if possible.

The uids only have to be unique over the course of a 24 hour period at a minimum, so 3 weeks until the timestamps repeat I think is plenty of leeway. As long as there are 2 or more updaters, a new timestamp should be getting grabbed every 1 or more seconds by a given updater. With how many updaters we currently have in either dev or prod, its around every 8 seconds. Also, I've had this branch running against the ARINC test server for a couple hours now and things seem to be stable.

hannahpurcell (approved):
Thanks to a Teams chat, I think this makes sense and is ready to go. Once merged, next step will be to verify with ARINC the logs look as expected.

PR #489, opened 2022-04-08
Tech spec: PA/ESS Monitoring Data Consumption

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Tech spec for PA/ESS health monitoring work](https://app.asana.com/0/1201786812162837/1202068421769296/f)

This is a tech spec for a proposed solution for two sources of monitoring logs/data that ARINC has agreed to provide the MBTA.

To view the tech spec in rendered rich text, click this icon in the top right of the view:
Screen Shot 2022-04-08 at 2 56 59 PM


docs/tech_specs/monitoring_data_consumption.md
```diff
@@ -0,0 +1,61 @@
...
+
+  2. ARINC sets up an endpoint for CTD to hit once per day around 2AM to receive a file containing the compiled logs. It would probably be good if the file is a zip to keep the size of data being transferred between servers small.
+
+  3. CTD sets up a recurring job which calls ARINC’s endpoint nightly to receive the file and dump it into an S3 bucket. Having the logs stored in S3 also makes it convenient if we ever need to dig up past log files for some reason.
+
+  4. CTD sets up an application that is responsible for pulling down the zip file from S3, unzipping it, and then streaming it into Splunk. There may be different options to explore on how to achieve this step. Splunk does not provide an easy way of uploading a CSV file through their API. They do provide a mechanism called a Universal Forwarder meant for forwarding large amounts of data into Splunk, though past attempts at setting these up have not been successful at the CTD. A "brute force" option could be to stream the CSV and `Logger.log()` the rows into Splunk, but the amount of requests would be close to 600,000 for one CSV for visual messages only.
```
paulswartz, line 33: I'd recommend something like Lambda or Batch for this, rather than a new ECS application, as doesn't need to be running all of the time. Lambda would be a Node or Python script, and Batch would be anything that can go into a Docker container.

You don't need to make a single request for each log line: they can be batched. See example 3 in the [Splunk documentation](https://docs.splunk.com/Documentation/Splunk/latest/Data/FormateventsforHTTPEventCollector) for the HTTP Event Collector. 

I'm not sure how many you could combine in a single request, but possibly it's all of them?

docs/tech_specs/monitoring_data_consumption.md
```diff
@@ -0,0 +1,61 @@
...
+
+The main reason for standing up a new application would be to avoid resource consumption on opstech3 and it's an effort to edge PA/ESS features towards AWS rather than increasing our dependence on on-prem servers.
+
+An alternative, potentially simpler option _could_ be to try streaming the CSV into Splunk directly from Realtime Signs and skip standing up an additional application since that will require some additional time and resources. We could still dump the zips or raw CSVs in S3 if we want to keep a record of what we receive from ARINC.
```
paulswartz, line 37: The other downside to this is that it might make the logs for Realtime Signs harder to work with, unless the logs would also be streaming from the PA/ESS (which it doesn't sound like).

docs/tech_specs/monitoring_data_consumption.md
```diff
@@ -0,0 +1,61 @@
...
+
+As a list of steps, the solution would involve the following in sequence: 
+
+  1. ARINC builds out an application that filters out unneeded logs and compiles the data that we outlined in our specification. With CTD having improved the usability of the `uid` field that we send in our messages to the ARINC headend server, this is achievable now.
+
+  2. ARINC sets up an endpoint for CTD to hit once per day around 2AM to receive a file containing the compiled logs. It would probably be good if the file is a zip to keep the size of data being transferred between servers small.
+
+  3. CTD sets up a recurring job which calls ARINC’s endpoint nightly to receive the file and dump it into an S3 bucket. Having the logs stored in S3 also makes it convenient if we ever need to dig up past log files for some reason.
```
paulswartz, line 31: This seems good. You could potentially use the new data platform S3 buckets, which we're planning to use for data such as this. (I'd still recommend starting with Splunk for the visualization)

PR #490, opened 2022-04-11
Add some more details around updating env variables

PaulJKim:
#### Summary of changes
Making edits to README.md to help clarify instructions on how to update environment variables.

This was brought up in a recent SEV when the idea of pointing Realtime Signs to the backup server was considered as an action to remediate the issue, but it was not well documented how to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Here's a general overview of the realtime_signs application which should be help
 
 ## Relevant environment variables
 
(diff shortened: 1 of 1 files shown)


PaulJKim:
@ianwestcott since you most recently had to interact with this document in order to understand how to update Realtime Signs to another server, do you have any feedback for how to make the documentation clearer?

PR #500, opened 2022-06-08
update build_release script to use elixir 1.12

PaulJKim:
The script used to build the release must also be updated to use Elixir 1.12

Fortunately, Elixir 1.12.3 is already present on opstech3 so we won't need to download it and put it there ourselves.
Screen Shot 2022-06-08 at 4 29 48 PM


losvedir (approved):
Oh, it's there already! Huh. Must have put it there for Trike? I don't remember. 🤔

PaulJKim:
Not complaining ¯\_(ツ)_/¯

PR #506, opened 2022-06-16
Set up recurring job to pull paess logs

PaulJKim:
This PR adds a CRON job which will request an endpoint that is hosted on ARINC's test server which serves a zip file containing log files with data on message latencies for every MBTA station with PA/ESS equipment for a given date. Firewall updates have been made with the help of ITD and the endpoint was tested through a web browser on opstech3.

After fetching the zip file, the job simply stores them in an S3 bucket which is meant to be the `mbta-ctd-dataplatform-dev-archive` bucket in a `paess` folder (Terraform changes were already made to allow permissions specific to this path in [this pr](https://github.com/mbta/devops/pull/416)). The logs will be used for regular auditing of message latencies across the system.

The [quantum](https://hexdocs.pm/quantum/readme.html) package is being used to enable the use of a CRON job. As of now, the plan is to have this job run nightly at 3am which is an hour or so after ARINC is planning to have their job scheduled to run which aggregates the message logs. The schedule is configurable though to allow for easy adjustment.

An endpoint was also added that can be used to manually run the job. The purpose of this is twofold.

1. Allow for easy test runs of the job.
2. Allow for manual re-runs in case the job fails to run for some reason. This could be for a number reasons such as the ARINC server being unreachable, their aggregation script failing to collect all the logs, opstech3 being down, etc.


lib/realtime_signs_web/controllers/monitoring_controller.ex
```diff
@@ -11,4 +11,10 @@ defmodule RealtimeSignsWeb.MonitoringController do
     monitor_device_uptime(nodes, timestamp)
     send_resp(conn, 200, "")
   end
+
+  def run_message_log_job(conn, %{"date" => date} = _params) do
```
losvedir, line 15: Maybe a `@doc` comment here at least explaining the format of the `date` expected?

config/runtime.exs
```diff
@@ -5,6 +5,22 @@ config :realtime_signs, RealtimeSignsWeb.Endpoint,
...
+  message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
+  message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER")
+
+config :realtime_signs, RealtimeSigns.Scheduler,
+  jobs: [
+    {
+      System.get_env("MESSAGE_LOG_CRON_SCHEDULE"),
+      {
+        RealtimeSigns.MessageLogJob, :get_and_store_logs,
+        [Date.utc_today() |> Date.add(-1) |> Date.to_string()]
```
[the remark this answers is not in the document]
PaulJKim: Ope duh, you're totally right. Thought I was being clever.

And yeah that should work! I had a similar approach set up before I tried this current one 😅

lib/jobs/message_log_job.ex
```diff
@@ -0,0 +1,35 @@
+defmodule RealtimeSigns.MessageLogJob do
+  require Logger
```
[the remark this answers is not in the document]
PaulJKim: Great recommendation, thank you! I'll keep that in mind in the future as well!

mix.lock
```diff
@@ -35,6 +41,11 @@
...
+  "telemetry": {:hex, :telemetry, "1.1.0", "a589817034a27eab11144ad24d5c0f9fab1f58173274b1e9bae7074af9cbee51", [:rebar3], [], "hexpm", "b727b2a1f75614774cff2d7565b64d0dfa5bd52ba517f16543e6fc7efcc0df48"},
+  "telemetry_registry": {:hex, :telemetry_registry, "0.3.0", "6768f151ea53fc0fbca70dbff5b20a8d663ee4e0c0b2ae589590e08658e76f1e", [:mix, :rebar3], [{:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "492e2adbc609f3e79ece7f29fec363a97a2c484ac78a83098535d6564781e917"},
+  "timex": {:hex, :timex, "3.7.8", "0e6e8bf7c0aba95f1e13204889b2446e7a5297b1c8e408f15ab58b2c8dc85f81", [:mix], [{:combine, "~> 0.10", [hex: :combine, repo: "hexpm", optional: false]}, {:gettext, "~> 0.10", [hex: :gettext, repo: "hexpm", optional: false]}, {:tzdata, "~> 1.1", [hex: :tzdata, repo: "hexpm", optional: false]}], "hexpm", "8f3b8edc5faab5205d69e5255a1d64a83b190bab7f16baa78aefcb897cf81435"},
```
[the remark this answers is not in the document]
PaulJKim: Yeah I think you're right, I think these got duplicated during a merge conflict resolution from the previous PR that set up the endpoint.

lib/jobs/message_log_job.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  def get_and_store_logs(date) do
+    Logger.info("Fetching message logs for #{date}")
+
+    case get_logs(date) do
+      {:ok, %{body: body}} ->
+        store_logs(body, s3_bucket(), s3_folder(), date)
```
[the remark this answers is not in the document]
PaulJKim: I added some handling logic in the `store_logs/4` method. Was wondering though, should we parse out the `status_code` and `body` from the response for the call to `aws_client.request()` and log those specifically, or do we think its probably fine to just log the whole response?

lib/realtime_signs.ex
```diff
@@ -67,6 +68,14 @@ defmodule RealtimeSigns do
     end
   end
 
+  defp consume_message_logs do
+    if Application.get_env(:realtime_signs, :message_log_zip_url) do
+      [RealtimeSigns.Scheduler]
```
losvedir, line 73: I'm kind of thinking the `Scheduler` should just be considered a new piece of the Realtime Signs infrastructure that always starts up. It might be useful for other things going forward. (I think some of the recurring `Engine` updates could be mediated a little more clearly through Quantum, for example.)

And then it would be the jobs themselves that are dynamically determined by the presence of an environment variable. So you'll always have the runner; it's just which jobs in which environment will it run?

losvedir (commented):
Couple more small comments but looking good!

lib/realtime_signs.ex
PaulJKim: Hmm yeah I can see how the scheduler could be useful other things! I've set up an approach where the job is now added dynamically. The thing I had some trouble with is that the job wasn't being added if I kept the Scheduler as a child process of the Supervisor since the Scheduler process wasn't running when I was calling my helper function that configures the job.

Curious as to what your thoughts would be for an approach to configuring and starting the job! What I have right now works, but I'm not certain how crucial it is to have the Scheduler running as a child process under the main Supervisor. Are you able to see a workaround where we can have the Scheduler first added as a child process and then configure/add the job after that point?

losvedir: Hmm, it could be that dynamically adding jobs is not really how it's made to be used, in which case I think probably the way you had it before is better, which was simpler and more straightforward to follow than what's there now, IMO.

What I had been envisioning was having the `RealtimeSigns.Scheduler` be unconditionally started by the supervision tree, and having the conditional be in `runtime.exs` where jobs were defined:

```elixir
scheduler_jobs =
  if System.get_env("MESSAGE_LOG_CRON_SCHEDULE") do
    [
      {
        System.get_env("MESSAGE_LOG_CRON_SCHEDULE"), 
        {RealtimeSigns.MessageLogJob, :get_and_store_logs, []}
      }
    ]
  else
    []
  end

config :realtime_signs, RealtimeSigns.Scheduler, jobs: scheduler_jobs
```

Does something like that not work?

PaulJKim: Ohh didn't think of binding the definition of the jobs to a variable like this. It works! Just updated the PR.

losvedir (approved):
Neat! 👍

PR #513, opened 2022-08-08
Union Square dest for D trains 

PaulJKim:
Specific pattern for the select D trains that are going to Union Sq. We need to add a new pattern because the usual default for D trains is North Station. We're using Science park because that's as far as we're predicting for these trains.


jfabi (approved):
Seems like this fits the bill, pending CI. ✅

PR #514, opened 2022-08-10
Enable logging of live PA equipment

PaulJKim:
ARINC is set up begin sending us online statuses of Live PA equipment so we should update the monitoring endpoint to parse these logs correctly and stream them into Splunk.

- DSP: Digital Signal Processor
- Amps: Amplifier
- Comrex: Comrex Live audio device


hannahpurcell (approved):
Cool! This looks great to me. Giving it a run will be the best test of it.

PR #516, opened 2022-08-17
Trains predicted to Science Park or Lechmere to use Union Sq

PaulJKim:
Any train being predicted out to Science Park or Lechmere will use Union Sq as the headway destination


jfabi (approved):
🟢 Looks pretty reasonable to me!

PR #519, opened 2022-09-01
Revert "Include platform info on JFK/UMass Mezzanine announcements"

PaulJKim:
Reverts mbta/realtime_signs#517

Reverting this as it seems to be causing intermittent outages on this specific sign. ARINC is investigating but in the meantime, we are going to revert the feature and monitor the sign to see if things stabilize.


hannahpurcell (approved):
Aye aye

PR #520, opened 2022-09-01
Remove hackney from realtime_signs

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Remove hackney from realtime_signs](https://app.asana.com/0/1189528452014899/1200505617474085/f)

This PR is a bit overdue, but it replaces the current main http client library being used `HTTPoison` with `Finch`. `HTTPoison` is a widely used HTTP Client library for Elixir built around hackney, but hackney has some limitations particularly when it comes to connection pooling. According to the Asana task, hackney has caused multiple failures for realtime-signs in the past. I suspect this may potentially be due to its error prone way of handling connections pools. More can be read on this [here ](https://elixirforum.com/t/mint-vs-finch-vs-gun-vs-tesla-vs-httpoison-etc/38588/11) in a forum post by the creator of `Finch`.

`Mojito`, another Elixir HTTP client library but built around `Mint` was the original suggestion to replace `HTTPoison`, but that library has since been deprecated and the author of `Mojito` recommends `Finch` as an alternative. `Finch` is specifically designed to handle very high throughput and more efficiently handle connection pooling and may suit realtime-sign's use case as a more stable http client.


PaulJKim:
According to the thread in [this GitHub issue](https://github.com/benoitc/hackney/issues/643) regarding the original bug motivating a move away from Hackney, it was resolved by the team managing Hackney with version 1.17 which we are using now. Therefore, I don't think there's any need to replace hackney at the moment and this PR can be closed out.

PR #521, opened 2022-09-01
Revert "Revert "Include platform info on JFK/UMass Mezzanine announcements""

PaulJKim:
Reverts mbta/realtime_signs#519

### DIFF
--- a/lib/content/message/predictions.ex
+++ b/lib/content/message/predictions.ex
@@ -162,7 +162,7 @@ defmodule Content.Message.Predictions do
                headsign_message,
                "#{duration_string}",
                width
-             ), 3},
+             ), 6},
             {headsign <> platform_message, 6}
           ]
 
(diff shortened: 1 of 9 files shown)


lib/content/message/predictions.ex
```diff
@@ -125,13 +146,34 @@ defmodule Content.Message.Predictions do
...
+                {headsign <> " (#{String.slice(platform_name, 0..0)})",
+                 " (#{platform_name} plat)"}
+
+          [
+            {Content.Utilities.width_padded_string(
+               headsign_message,
+               "#{duration_string}",
+               width
+             ), 3},
+            {headsign <> platform_message, 6}
```
PaulJKim, line 166: Revert of the revert, except with this value increased to 6.

The reason for this increase of the page duration is because ARINC suspects that the original feature was causing the sign to go offline because there was too much data going to the sign and its buffer was overflowing. Decreasing the frequency of the paging should help this.

hannahpurcell (approved):
let's give it a go!

PR #532, opened 2022-10-13
Realtime Signs Dockerfile

PaulJKim:
Just adding the dockerfile. 

Image successfully builds locally.


config/dev.exs
```diff
@@ -1,11 +1,4 @@
...
 
 config :logger,
-  backends: [{Logger.Backend.Splunk, :splunk}, :console]
-
-config :logger, :splunk,
-  connector: Logger.Backend.Splunk.Output.Http,
-  host: 'https://http-inputs-mbta.splunkcloud.com/services/collector/event',
-  token: {:system, "SIGNS_SPLUNK_TOKEN"},
-  format: "$dateT$time [$level]$levelpad $metadata$message\n",
-  metadata: [:request_id]
```
PaulJKim, line 11: This doesn't get used on Opstech3 since the "dev" instance is still a prod release but with a different config so we can probably just delete this one anyway. And when developing locally, we don't want to be logging to Splunk. The prod config I will remove/address a little later down the line for this migration.

paulswartz (approved):
:cake: can confirm this builds! It does fail to start, but I think that's expected at this stage.

PR #533, opened 2022-10-17
Workflows for deploying to on prem servers

PaulJKim:
Adding github workflows to deploy to dev and prod on-prem clusters.

Also removing splunk logger backend from the prod config.

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -1,0 +1,22 @@
+name: Deploy to Prod
+
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy:
+    name: Deploy
+    uses: mbta/workflows/.github/workflows/deploy-on-prem.yml@main
+    with:
+      app-name: realtime-signs
+      environment: prod
+      on-prem-cluster: hsctd-prod-managers
+      splunk-index: realtime-signs-prod
+      task-cpu: 1
+      task-memory: 512M
+      task-port: 80
+    secrets:
+      aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      docker-repo: ${{ secrets.DOCKER_REPO }}
+      slack-webhook: ${{ secrets.SLACK_WEBHOOK }}
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -12,8 +12,8 @@ jobs:
       environment: prod
       on-prem-cluster: hsctd-prod-managers
       splunk-index: realtime-signs-prod
-      task-cpu: 1
-      task-memory: 512M
+      task-cpu: .5
+      task-memory: 1024M
       task-port: 80
     secrets:
       aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -35,10 +35,10 @@ if config_env() == :prod and splunk_token != "" do
   config :logger, backends: [Logger.Backend.Splunk, :console]
 
   config :logger, :splunk,
-  connector: Logger.Backend.Splunk.Output.Http,
-  host: 'https://http-inputs-mbta.splunkcloud.com/services/collector/event',
-  token: splunk_token,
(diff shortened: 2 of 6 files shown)


.github/workflows/prod.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  deploy:
+    name: Deploy
+    uses: mbta/workflows/.github/workflows/deploy-on-prem.yml@main
+    with:
+      app-name: realtime-signs
+      environment: prod
+      on-prem-cluster: hsctd-prod-managers
+      splunk-index: realtime-signs-prod
+      task-cpu: 1
```
PaulJKim, line 15: I'm wondering if it's worth increasing the available CPU for prod at all. There's a fair amount of concurrency happening in realtime-signs, namely the GenServers for each sign and the pool of HTTP updaters that concurrently send requests to ARINC's server. If my understanding of the BEAM is correct, I think we would only reap performance benefits of the concurrency framework if there's more than 1 CPU core available.

config/runtime.exs
```diff
@@ -27,3 +27,18 @@ config :realtime_signs, RealtimeSigns.Scheduler, jobs: scheduler_jobs
...
+splunk_token = System.get_env("PROD_SIGNS_SPLUNK_TOKEN", "")
+
+if config_env() == :prod and splunk_token != "" do
+  config :logger, backends: [Logger.Backend.Splunk, :console]
+
+  config :logger, :splunk,
+  connector: Logger.Backend.Splunk.Output.Http,
+  host: 'https://http-inputs-mbta.splunkcloud.com/services/collector/event',
+  token: splunk_token,
+  format: "$dateT$time [$level]$levelpad node=$node $metadata$message\n",
```
paulswartz, line 41: question: is the `:console` logger configured to use the format as well?

.github/workflows/prod.yml
paulswartz: This value isn't really whether there are multiple CPUs available to the task. The [Docker documentation](https://docs.docker.com/config/containers/resource_constraints/) talks more about it, but it's a limit to the number of CPU shares that the task can use without being throttled. I don't think that `realtime-signs` is doing a lot of CPU-bound work, but you can see how well it works by running the container locally with `docker run ... -c 1.5 -m "512M" realtime_signs`.

config/runtime.exs
PaulJKim: Missed that, thank you for the catch!

.github/workflows/prod.yml
PaulJKim: I see, so it seems like `task-cpu` affects the containers access to CPU cycles, which doesn't necessarily have to do with the containers access to available cores.

running the container locally, the application appears to work normally so I think we can keep `task-cpu` set to 1

paulswartz: 👍 It might not even need 1 CPU: a lot of our smaller applications (commuter-rail-boarding, parking-tweets, api-checker) are only set to 0.25 of a CPU. You could try it out locally with less and see how it works.

PaulJKim: I've fiddled around with it a bit and it does seem to run fine with .25 or .5 of a CPU. The container itself seems a little less responsive with .25 CPU (slower start up and stopping) but I think that's not that surprising. 

I think the memory might need to be upped though. With 512Mb of memory, the application eventually crashes after seemingly running out of memory on the heap. The error message is `eheap_alloc: Cannot allocate 85562808 bytes of memory (of type "heap")`

When I run the image with 1024Mb of memory though, it doesn't crash. I don't think I fully understand what's happening here though. It seems like it's possible this could point to a memory leak somewhere in realtime-signs, but again not sure.

paulswartz: I did notice that for some of the other applications as well, that they needed more memory than I expected.

PaulJKim: oh, curious.. well, I updated the values though to be .5 of a CPU and 1024Mb of memory. What do the available resources look like on the actual windows servers we're deploying to? 1Gb of memory for realtime-signs does seem like a lot 🤔

paulswartz: for the non-production environment, currently it's 1 CPU and 8GB RAM. for the production environment, 3 CPU and 24GB RAM. we're in-progress to get an additional 2CPU and 16GB RAM for each environment as well.

paulswartz (approved):
🍰

PR #534, opened 2022-10-17
Add a dry run mode

PaulJKim:
Overrides default configs and uses the `PaEss.Logger` module as the sign updater module if an env variable `DRY_RUN` is set. 

This module has the same interface as `MessageQueue` which is used when we want to actually send http requests to the headend server. `MessageQueue` is the default for normal deploys to either dev or prod, but we can use this dry run feature when we are testing out deploying to the new on prem servers and don't want to risk potentially overwhelming the PAESS system with too many messages.

### DIFF
--- a/lib/pa_ess/logger.ex
+++ b/lib/pa_ess/logger.ex
@@ -1,6 +1,6 @@
 defmodule PaEss.Logger do
   @moduledoc """
-  A behaviour that signs can use to append their updates to a file,
+  A behaviour that signs can use to log their updates
   rather than sending it to the Pa/Ess HTTP server.
   """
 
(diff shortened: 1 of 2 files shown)


lib/pa_ess/logger.ex
```diff
@@ -24,8 +24,6 @@ defmodule PaEss.Logger do
       "#{start_secs}"
     ]
 
-    File.mkdir("log")
-    File.write!("log/pa_ess_updates.log", line ++ ["\n"], [:append])
```
PaulJKim, line 28: I don't think we'll need realtime signs to log to a local file. This would just log the updates to a file when running locally to help with debugging, but I found that it's not super useful.

cmaddox5 (approved):

PR #536, opened 2022-10-19
Use awscli credentials

PaulJKim:
Use awscli credentials for `ExAws`


paulswartz (approved):
Looks good, but I would try deploying it to make sure there aren't any other issues.

PR #542, opened 2022-11-01
Message latency reporting job

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Outline the message latency evaluation steps & outputs](https://app.asana.com/0/1201786812162837/1203092194190502/f)

This job is responsible for generating daily ARINC message latency reports based on a provided start_date and days_to_analyze.

### Params:
  * start_date: The date from which the job should begin generating reports. The job
  will work backwards from this date. Defaults to Date.utc_today() - 1 because those
  will be the latest logs that we have when this job runs on its CRON schedule.
  * days_to_analyze: The number of days to go back. This value will default to 7 because
  we plan to schedule the job to run on a weekly basis.

The job downloads the zip file containing message latency log files for each station and calculates the 95th and 99th percentiles along with the total row count for each one. It then aggregates into a CSV file the stats both for the combined logs from every station as well as on a per station basis for a given day. The job will then store this file in S3.

Example generated CSV:
[2022-10-31-message_latency(10).csv](https://github.com/mbta/realtime_signs/files/9914366/2022-10-31-message_latency.10.csv)

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -28,7 +28,7 @@ message_log_report_job =
     [
       {
         System.get_env("MESSAGE_LOG_REPORT_CRON_SCHEDULE"),
-        {Jobs.MessageLatencyReport, :generate_message_latency_report, []}
+        {Jobs.MessageLatencyReport, :generate_message_latency_reports, []}
       }
     ]
   else
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,9 +8,10 @@ config :realtime_signs, RealtimeSignsWeb.Endpoint,
 config :realtime_signs,
   message_log_zip_url: System.get_env("MESSAGE_LOG_ZIP_URL"),
   message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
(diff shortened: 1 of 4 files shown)


PaulJKim:
@cmaddox5 I don't calculate the average. Instead it's the value/row that 95% of the messages have a latency lower than. So for the example value of 1.582 seconds, it tells us that 95% of the messages took less than that time. And likewise for the 99th percentile.

The way I calculate it is more based on sorting and then finding the index where the bottom 95% begins, if that makes sense. So like [here](https://github.com/mbta/realtime_signs/pull/542/files#diff-55c8accadee4f46e20014e50009ed1786e9b8c57ed455ca8803a78dc3cb0d63fR186) what I'm trying to do is take the indices, find the index that represents 95% of the list, then use that to get the row where I'll take the `seconds` value to include in the final report.

Hopefully that makes sense! I used [this medium article](https://medium.com/@ankur_anand/an-in-depth-introduction-to-99-percentile-for-programmers-22e83a00caf) as a guideline for how to approach the math if it's helpful.

cmaddox5:
Ahh, makes total sense. Thanks for clarifying!

cmaddox5 (approved):

PR #544, opened 2022-11-04
Update headway destination values for E, GLX stations, and North Station

PaulJKim:
These changes only pertain to default values for signs set to headway mode 
- Headway mode for E Line only stations will use `Lechmere` as headway destination (Symphony and Prudential)
- Headway mode for Union Sq will use `Riverside` as headway destination because primarily D trains run out of there now
- Add `Inbound` and `Outbound` as possible values for headway destinations
- Green Line North Station will use `Inbound/Outbound`
- Science will use `Inbound/Outbound`
- Lechmere will use `Inbound/Outbound`

PIO Eryk Rotondo was made aware of and is ok with the use of Inbound/Outbound at these few stations https://mbta.slack.com/archives/C049FR1UQCV/p1667582256334679

Screenshots taken in dev:
Screen Shot 2022-11-04 at 3 04 42 PM

Screen Shot 2022-11-04 at 3 04 17 PM

### DIFF
--- a/lib/content/utilities.ex
+++ b/lib/content/utilities.ex
@@ -69,7 +69,7 @@ defmodule Content.Utilities do
   def destination_for_prediction("Green-E", 0, _), do: {:ok, :heath_street}
   def destination_for_prediction("Green-B", 1, _), do: {:ok, :government_center}
   def destination_for_prediction("Green-C", 1, _), do: {:ok, :government_center}
-  def destination_for_prediction("Green-D", 1, _), do: {:ok, :north_station}
+  def destination_for_prediction("Green-D", 1, _), do: {:ok, :union_square}
   def destination_for_prediction("Green-E", 1, _), do: {:ok, :lechmere}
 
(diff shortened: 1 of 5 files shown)


lib/content/utilities.ex
```diff
@@ -69,7 +69,7 @@ defmodule Content.Utilities do
   def destination_for_prediction("Green-E", 0, _), do: {:ok, :heath_street}
   def destination_for_prediction("Green-B", 1, _), do: {:ok, :government_center}
   def destination_for_prediction("Green-C", 1, _), do: {:ok, :government_center}
-  def destination_for_prediction("Green-D", 1, _), do: {:ok, :north_station}
+  def destination_for_prediction("Green-D", 1, _), do: {:ok, :union_square}
```
PaulJKim, line 72: Snuck in this small change here. This clause is mainly a catch-all for D Line predictions and needed to be updated.

jzimbel-mbta (approved):
Apologies for the slow review!

PR #545, opened 2022-11-07
Tweak message latency job to filter out bad dates

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Automate message latency evaluation](https://app.asana.com/0/1201786812162837/1203092194190502/f)

ARINC's message latency logs sometimes contain logs from dates that should be out-of-scope. They're working on improving this but it's unclear when it'll be done. We can go ahead and filter the "bad" dates out on our end.

On top of this, the formatting of their dates is a bit inconsistent as well. Similarly, we've called it out and they are working on it, but they're not able to provide a clear timeline on when it'll be fixed. In the meantime, we can fit our logic to deal with the varying formats.

### DIFF
--- a/lib/jobs/message_latency_report.ex
+++ b/lib/jobs/message_latency_report.ex
@@ -51,7 +51,10 @@ defmodule Jobs.MessageLatencyReport do
           :os.cmd(:"type #{unzipped_directory}\\*.csv > #{unzipped_directory}\\all.csv")
       end
 
-      [get_csv_row("#{unzipped_directory}/all.csv") | Enum.map(files, &get_csv_row/1)]
+      [
+        get_csv_row("#{unzipped_directory}/all.csv", date)
+        | Enum.map(files, &get_csv_row(&1, date))
+      ]
       |> format_csv_data()
       |> put_csv_in_s3(date)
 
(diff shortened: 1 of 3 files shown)


lib/jobs/message_latency_report.ex
```diff
@@ -123,15 +135,39 @@ defmodule Jobs.MessageLatencyReport do
         station: station
       ]
     end)
-    |> Enum.sort_by(fn row -> row[:seconds] end)
-    |> Enum.with_index()
```
PaulJKim, line 127: We don't need to sort + index the rows until after we've filtered out the irrelevant dates so I moved this into `get_stats/2`

lib/jobs/message_latency_report.ex
```diff
@@ -123,15 +135,39 @@ defmodule Jobs.MessageLatencyReport do
...
+  end
+
+  defp filter_by_date(row, date) do
```
PaulJKim, line 140: I admit, this helper function is a little hacky, but it seems like Elixir doesn't have very rich support for parsing datetimes and this gets the job done.

lib/jobs/message_latency_report.ex
```diff
@@ -123,15 +135,39 @@ defmodule Jobs.MessageLatencyReport do
...
+    # Get the day of month (0-31) from the date for which the logic is running
+    [_, _, current_day] = date |> String.split("-")
```
cmaddox5, line 142: It might look a little cleaner to convert `date` to a `Date` right off the bat and access the day that way. You are already converting it a couple of lines down so shouldn't throw too much off. This works though so not a blocking suggestion.

cmaddox5 (approved):
Sorry for the delay! Looks good! Just one suggestion, but nothing blocking an approval.

lib/jobs/message_latency_report.ex
PaulJKim: Oh good point! I like that suggestion and just tweaked to code. Cleaner indeed!

PR #549, opened 2022-11-14
GLX Branch 2

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Prepare for GLX Medford branch](https://app.asana.com/0/1201786812162837/1203228566813180/f)

This PR contains changes that are necessary to enable realtime-signs to send messages to the PA/ESS hardware at the new GLX stations.

Note: This PR should **not** be deployed to prod until the night before revenue service begins on branch 2

There's a lot of small changes in many different places so here is an outline of the changes made in this PR to help:
- `lib/pa_ess.ex` - add `:medford_tufts` as a valid value as a destination
- `lib/content/audio/train_is_arriving.ex`, `lib/content/audio/vehicles_to_destination.ex`, and `lib/content/audio/train_is_boarding.ex`, `lib/pa_ess/utilities.ex` line 197 - relevant audio additions. 
- `lib/content/utilities.ex` - predictions going out to College Ave will use `Medford` as headway destination
- `lib/pa_ess/utilities.ex` - `Medford Tufts` can now be parsed from the source config (signs.json) and headway mode will use `Medfd/Tufts`
- `priv/signs.json` - Configure data sources for new stations and assign audio/visual zones
- `priv/stops.json` - Allow alerts to be applied to new station countdown clocks


lib/content/audio/train_is_boarding.ex
```diff
@@ -53,7 +53,8 @@ defmodule Content.Audio.TrainIsBoarding do
                 :government_center,
                 :park_st,
                 :kenmore,
-                :union_square
+                :union_square,
+                :medford_tufts
```
PaulJKim, line 57: I believe there are very few, if any, cases where we announce boarding trains for Green line so this change may not be necessary. But I decided to include it for good measure.

jzimbel-mbta (approved):
LGTM! There are one or two places in the diff where you used the string "Medford Tufts" where I'd have expected "Medford/Tufts", but I'm guessing that's because of the limitations of what characters the PA/ESS screens can display?

PaulJKim:
Hm that's a fair point! The countdown clocks actually can display slashes. I think for the values in `signs.json` I did that subconsciously because the text `Heath Street` was nearby. The `Medford Tufts` string in `lib/pa_ess/utilities.ex` would rarely, if ever, get used and it would automatically be matched and replaced with `Medfd/Tufts` anyhow before it got sent to ARINC.

I think for consistency's sake though, it might be worth changing these to Medford/Tufts. I'll make that change and test the effects with a branch deploy before merging/deploying to dev!

PR #553, opened 2023-01-03
Migrate runtime config to runtime.exs

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Chore: Migrate to using runtime.exs for runtime configuration values](https://app.asana.com/0/1201753694073608/1203627766980875/f)

Since Elixir 1.11, we have the option of having a file named `runtime.exs` as part of our config to contain runtime config values. Realtime Signs was initially set up before Elixir 1.11 so there was a manual solution that was being used. But since Elixir is able to do the work for us natively, we should take advantage of that.

The runtime config allows us to change config values with a restart rather than having to recompile everything. Most of the time, we're probably more likely to fully redeploy the application anyway, but in certain situations the quicker change to the underlying env variables + restarting can be advantageous.


config/runtime.exs
```diff
@@ -52,7 +63,7 @@ if config_env() == :prod and splunk_token != "" do
     connector: Logger.Backend.Splunk.Output.Http,
     host: 'https://http-inputs-mbta.splunkcloud.com/services/collector/event',
     token: splunk_token,
-    format: "$dateT$time [$level]$levelpad node=$node $metadata$message\n",
+    format: "$dateT$time [$level] node=$node $metadata$message\n",
```
PaulJKim, line 66: `$levelpad` was deprecated with this PR https://github.com/elixir-lang/elixir/pull/11633

config/config.exs
```diff
@@ -24,24 +24,14 @@ import Config
...
   scheduled_headway_requester: Headway.Request,
   headway_calculator: Headway.HeadwayDisplay,
   external_config_getter: ExternalConfig.Local,
   aws_client: ExAws,
   s3_client: ExAws.S3,
-  s3_bucket: nil,
-  s3_path: nil,
-  api_v3_key: nil,
-  api_v3_url: "https://api-dev-green.mbtace.com",
   number_of_http_updaters: 4,
```
panentheos, line 35: Since this is now written in the runtime config, we should remove it from here.

panentheos (approved):
See comment, then looks good. Tested by running locally and setting some environment variables.

config/config.exs
PaulJKim: Ah good catch! So this one we actually do need to keep for this PR because of another bug. Basically, in `http_updater.ex`, this config is actually read at compile time rather than runtime, even though it is treated in `realtime_signs.ex` like a runtime config value so it still needs a value for `number_of_http_updaters` at compile time.

I have a PR almost ready for making that change at #554 , but I still need to remove this config value, so I'll do it over in that PR!

PR #554, opened 2023-01-04
Calculate max send rate and avg wait time at runtime

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Chore: Revisit max send rate](https://app.asana.com/0/1201753694073608/1203627766980881/f)

In Elixir, module attributes are evaluated at compile time. This means that `@max_send_rate_per_sec` and `@avg_ms_between_sends` were being calculated based on compile time values which are just the defaults. The config value that we actually want to use gets set at runtime instead, which means that those values were not being utilized.

This PR moves the calculation of these values to the initialization step of the http_updater GenServers so that we will evaluate them at runtime.


PaulJKim:
I seem to have messed up my branch while trying to rebase 😬 remaking the branch and closing this PR

PR #555, opened 2023-01-05
Calculate max send rate and avg wait time at runtime

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Chore: Revisit max send rate](https://app.asana.com/0/1201753694073608/1203627766980881/f)

In Elixir, module attributes are evaluated at compile time. This means that `@max_send_rate_per_sec` and `@avg_ms_between_sends` were being calculated based on compile time values which are just the defaults. The config value that we actually want to use gets set at runtime instead, which means that those values were not being utilized.

This PR moves the calculation of these values to the initialization step of the http_updater GenServers so that we will evaluate them at runtime.

### DIFF
--- a/lib/pa_ess/http_updater.ex
+++ b/lib/pa_ess/http_updater.ex
@@ -40,7 +40,6 @@ defmodule PaEss.HttpUpdater do
       http_poster: http_poster,
       queue_mod: queue_mod,
       updater_index: opts[:updater_index],
-      max_send_rate_per_sec: max_send_rate_per_sec,
       avg_ms_between_sends: avg_ms_between_sends
     )
   end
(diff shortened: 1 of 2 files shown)


PaulJKim:
I seem to have messed up my branch while trying to rebase 😬 remaking the branch and closing this PR

lib/pa_ess/http_updater.ex
```diff
@@ -34,11 +28,20 @@ defmodule PaEss.HttpUpdater do
...
+
+    avg_ms_between_sends = round(1000 / max_send_rate_per_sec)
+
     GenServer.start_link(
       __MODULE__,
       http_poster: http_poster,
       queue_mod: queue_mod,
-      updater_index: opts[:updater_index]
+      updater_index: opts[:updater_index],
+      max_send_rate_per_sec: max_send_rate_per_sec,
```
panentheos, line 43: nit: It looks like this value doesn't get used later on, so we don't technically need to store it in state

panentheos (approved):
Looks good. Tested by running locally.

lib/pa_ess/http_updater.ex
PaulJKim: Ah yeah that's a good point, will remove that then merge

PR #556, opened 2023-01-05
allow controlling network calls via environment variables

panentheos:
#### Summary of changes
**Asana Ticket:** [Reduce friction when developing locally with signs_ui](https://app.asana.com/0/1201753694073608/1203640960063074/f)

This makes the network calls to the headend server and signs_ui conditional on their respective environment variables, so it's possible to run locally against a local signs_ui instance without changing the code. Tested by running in several different configurations locally.

#### Reviewer Checklist


config/runtime.exs
```diff
@@ -6,8 +6,8 @@ config :realtime_signs, RealtimeSignsWeb.Endpoint,
   server: true
 
 config :realtime_signs,
-  sign_head_end_host: System.get_env("SIGN_HEAD_END_HOST", "127.0.0.1"),
-  sign_ui_url: System.get_env("SIGN_UI_URL", "signs-dev.mbtace.com"),
+  sign_head_end_host: System.get_env("SIGN_HEAD_END_HOST", Application.get_env(:realtime_signs, :sign_head_end_host)),
+  sign_ui_url: System.get_env("SIGN_UI_URL", Application.get_env(:realtime_signs, :sign_ui_url)),
```
PaulJKim, line 10: Is the thought here to have `sign_head_end_host` and `sign_ui_url` default to `nil` whenever they aren't set in an env variable? If so, we can omit the call to `Application.get_env/2` and just not include a default value since `System.get_env/1` will automatically return `nil` if it doesn't find the env var that was requested.

PaulJKim (approved):
Mostly LGTM! Tested locally and it works, although I had one comment that might clean up the code a little bit.

config/runtime.exs
panentheos: That `get_env` is there for test mode, where we want network calls to be "enabled", even though the actual HTTP module is mocked. The test config sets up dummy hostnames, and this code makes sure that's retained instead of overwritten by the lack of an environment variable.

PaulJKim: Ahh gotcha that makes sense!

PR #557, opened 2023-01-06
prefer showing distinct destinations if a stop has multiple

panentheos:
#### Summary of changes
**Asana Ticket:** [Establish a fixed order of RL SB predictions for RL SB zones](https://app.asana.com/0/1201786812162837/1203617720856117/f)

This modifies the prediction logic to prefer showing distinct destinations when the prediction list has multiple destinations. This required moving the logic further down in the pipeline, since we want to choose based on the aggregated message `destination`, not the granular `destination_stop_id` that's in the raw prediction.

Tested by hardcoding some duplicate predictions and verifying that it treats them correctly.

### DIFF
--- a/lib/signs/utilities/predictions.ex
+++ b/lib/signs/utilities/predictions.ex
@@ -45,7 +45,6 @@ defmodule Signs.Utilities.Predictions do
          end
        end, prediction.seconds_until_departure, prediction.seconds_until_arrival}
     end)
-    |> Enum.take(2)
     |> Enum.map(fn {source, prediction} ->
       cond do
         stopped_train?(prediction) ->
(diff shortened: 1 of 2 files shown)


PaulJKim (approved):
This looks great! Tested locally with local signs UI and compared with dev/prod signs UI and things looked right.

I actually think this might be a preferable approach to apply to mezzanine signs over the dedicated line approach that we take on most of them now. I see two advantages:
1. Earliest prediction first is preserved, which is not the case when we dedicate a line to a specific direction as pictured below
Screenshot 2023-01-06 at 1 28 08 PM
2. It's a little more dynamic and takes better advantage of the real estate we have available on the signs. Namely, if there aren't predictions for a particular direction for some reason, we'll start showing an additional prediction for the other direction, offering just a bit more information to riders. Currently, the top or bottom line would just go blank which I think is no better than just showing two predictions in one direction.

Let me know what you think about the second point! If we agree, we can pitch it to Kevin and if he agrees, we can make the config changes in a follow-up task.

PR #558, opened 2023-01-09
update README with local dev setup instructions

panentheos:
#### Summary of changes
**Asana Ticket:** [Update README with info about running locally with signs_ui](https://app.asana.com/0/1201753694073608/1203659128307178/f)

This updates the documentation with the new method for running locally against a local signs_ui.


PaulJKim (approved):
Thank you for adding this!

PR #560, opened 2023-01-18
Filter out JFK predictions > 5 min

PaulJKim:
Upon revisiting JFK/UMass predictions and doing a finer-grained analysis, we found that platform predictions at JFK/UMass tend to get less accurate after around the 5 minute threshold. This adjustment is a response to that research.

When the next Alewife train is more than 5 minutes away, we will use the "Platform TBD" message to encourage riders to wait in the mezzanine until the next train is closer.

### DIFF
--- a/lib/content/audio/predictions.ex
+++ b/lib/content/audio/predictions.ex
@@ -88,7 +88,9 @@ defmodule Content.Audio.Predictions do
           minutes: predictions.minutes,
           verb: if(src.terminal?, do: :departs, else: :arrives),
           track_number: Content.Utilities.stop_track_number(predictions.stop_id),
-          platform: src.platform
+          platform: src.platform,
+          station_code: predictions.station_code,
+          zone: predictions.zone
         }
     end
   end
--- a/lib/content/audio/predictions.ex
+++ b/lib/content/audio/predictions.ex
@@ -66,7 +66,9 @@ defmodule Content.Audio.Predictions do
           minutes: 1,
           verb: if(src.terminal?, do: :departs, else: :arrives),
           track_number: Content.Utilities.stop_track_number(predictions.stop_id),
-          platform: src.platform
+          platform: src.platform,
+          station_code: predictions.station_code,
+          zone: predictions.zone
         }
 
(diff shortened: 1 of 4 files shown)


lib/content/message/predictions.ex
```diff
@@ -151,11 +151,11 @@ defmodule Content.Message.Predictions do
           platform_name = Content.Utilities.stop_platform_name(stop_id)
 
           {headsign_message, platform_message} =
-            if minutes == :max_time,
-              do: {headsign, " (Platform TBD)"},
-              else:
-                {headsign <> " (#{String.slice(platform_name, 0..0)})",
-                 " (#{platform_name} plat)"}
+            unless minutes == :max_time or (is_integer(minutes) and minutes > 5) do
```
panentheos, line 154: nit: This logic looks right, but I find `unless` with `else` hard to read due to the double negative. Perhaps we should stick with `if`, and maybe invert the logic if it's preferable to have the typical case first.

lib/content/audio/predictions.ex
```diff
@@ -88,7 +90,9 @@ defmodule Content.Audio.Predictions do
           minutes: predictions.minutes,
           verb: if(src.terminal?, do: :departs, else: :arrives),
           track_number: Content.Utilities.stop_track_number(predictions.stop_id),
-          platform: src.platform
+          platform: src.platform,
+          station_code: predictions.station_code,
+          zone: predictions.zone
```
panentheos, line 95: Note for onlookers, since this wasn't obvious to me at first: Previously the check in the audio code path wasn't even getting hit on <20 min predictions, because we weren't passing this data, but that didn't matter because the check was only relevant for 20+ min anyway. Now that it's lower, we need to feed this data through so the check gets hit all the time.

panentheos (approved):
See comment, then looks good. I did not do any additional local testing.

lib/content/message/predictions.ex
PaulJKim: That's fair, I agree that `unless` can be difficult to read at a first pass. And I'm cool with leaving the logic as is to keep things simple!

PR #562, opened 2023-01-18
Add special casing for Kenmore inbound announcements

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Update audio announcements at Kenmore](https://app.asana.com/0/1201753694073608/1203674455933121/f)

This special case logic is aimed to help improve clarity for blind/low-vision riders at Kenmore, which has a unique platform setup and can sometimes be confusing to riders. The new logic will ensure that whenever we're announcing the next or following train at Kenmore that we include the branch letter along with it. So for example, we would announce something like "The next C train to Gov't center arrives in 2 minutes. The following D train to Union Square arrives in 4 minutes". We already announce the branch letter in the westbound direction so there would be no change there.

Example of an audio request announcing the next C train to Gov't Center
- `MsgType=Canned&uid=563932713&mid=117&var=501%2C21000%2C537%2C21000%2C507%2C21000%2C4061%2C21000%2C503%2C21000%2C504%2C21000%2C5007%2C21000%2C505&typ=1&sta=GKEN001000&pri=5&tim=60`

Example of an audio request announcing a following B train to Gov't Center
- `MsgType=Canned&uid=563986502&mid=117&var=667%2C21000%2C536%2C21000%2C507%2C21000%2C4061%2C21000%2C503%2C21000%2C504%2C21000%2C5006%2C21000%2C505&typ=1&sta=GKEN000010&pri=5&tim=60`

### DIFF
--- a/lib/content/utilities.ex
+++ b/lib/content/utilities.ex
@@ -97,4 +97,10 @@ defmodule Content.Utilities do
   def route_and_destination_branch_letter("Green-D", :reservoir), do: :d
   def route_and_destination_branch_letter("Green-E", :heath_street), do: :e
   def route_and_destination_branch_letter(_route_id, _destination), do: nil
+
+  @spec route_branch_letter(String.t()) :: green_line_branch()
+  def route_branch_letter("Green-B"), do: :b
+  def route_branch_letter("Green-C"), do: :c
+  def route_branch_letter("Green-D"), do: :d
+  def route_branch_letter("Green-E"), do: :e
 end
--- a/lib/content/audio/following_train.ex
+++ b/lib/content/audio/following_train.ex
@@ -3,8 +3,8 @@ defmodule Content.Audio.FollowingTrain do
   The following train to [destination] arrives in [n] minutes.
   """
 
-  @enforce_keys [:destination, :route_id, :verb, :minutes, :station_code]
-  defstruct @enforce_keys
+  @enforce_keys [:destination, :route_id, :verb, :minutes]
+  defstruct @enforce_keys ++ [station_code: nil]
 
   @type verb :: :arrives | :departs
 
@@ -13,7 +13,7 @@ defmodule Content.Audio.FollowingTrain do
           route_id: String.t(),
           verb: verb(),
           minutes: integer(),
-          station_code: String.t()
+          station_code: String.t() | nil
         }
 
   require Logger
--- a/lib/content/audio/following_train.ex
+++ b/lib/content/audio/following_train.ex
@@ -3,7 +3,7 @@ defmodule Content.Audio.FollowingTrain do
   The following train to [destination] arrives in [n] minutes.
   """
 
-  @enforce_keys [:destination, :route_id, :verb, :minutes]
+  @enforce_keys [:destination, :route_id, :verb, :minutes, :station_code]
   defstruct @enforce_keys
 
   @type verb :: :arrives | :departs
@@ -12,7 +12,8 @@ defmodule Content.Audio.FollowingTrain do
           destination: PaEss.destination(),
           route_id: String.t(),
           verb: verb(),
-          minutes: integer()
+          minutes: integer(),
+          station_code: String.t()
         }
 
   require Logger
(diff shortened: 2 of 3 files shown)


lib/content/utilities.ex
```diff
@@ -97,4 +97,10 @@ defmodule Content.Utilities do
   def route_and_destination_branch_letter("Green-D", :reservoir), do: :d
   def route_and_destination_branch_letter("Green-E", :heath_street), do: :e
   def route_and_destination_branch_letter(_route_id, _destination), do: nil
+
+  @spec route_branch_letter(String.t()) :: green_line_branch()
+  def route_branch_letter("Green-B"), do: :b
+  def route_branch_letter("Green-C"), do: :c
+  def route_branch_letter("Green-D"), do: :d
+  def route_branch_letter("Green-E"), do: :e
```
panentheos, line 105: Thinking out loud: There's some slight duplication of logic here, but I'm not sure if it's worth worrying about. Fundamentally there are two calculations here: "Should we read out the branch letter?" and "What is the branch letter for this route?" Previously `route_and_destination_branch_letter` was doing both jobs, but now the logic is expanded such that it would be increasingly awkward to keep things bundled together like that.

I like `route_branch_letter`, it's a pure representation of that second computation, so if anything I think `route_and_destination_branch_letter` could be changed, but that would require addressing the third call site in `train_is_boarding.ex`. Presumably we don't *need* the expanded logic there because we don't make Boarding announcements at that station (?), but conceptually maybe we should use uniform logic, even if it doesn't apply right now. However, that would require wiring `station_code` in there, where it currently isn't.

Thoughts?

PaulJKim: Hmm great points. I agree that the duplication of logic here and awkward bundling of the logic isn't preferable. Also agreed that `route_branch_letter` is good and I think separating out the logic to answer "should we read out the branch letter?" is a step in the right direction. It'll be easier to maintain and the way it is now sort of obfuscates how we make that decision at all.

Honestly, I'm not really liking having to hardcode station codes in various places (with the exception of `RJFK` because it's such an extremely specific case). I'm considering an alternative approach that involves adding to the source configuration schema. What if we added an additional boolean field like `announce_branch_letter` that we then pass that down to the audio structs and use that to decide whether to announce branch names or not? Then we can just eliminate `route_and_destination_branch_letter`, check the value of the new flag, and stick to `route_branch_letter` alone when we need to know what branch letter to use.

PaulJKim: That approach would touch more places, but I think it might make for a cleaner and easier to maintain solution. It also should make it easier if we ever want to start announcing inbound branch letters at other stations.

panentheos: The disadvantage of using the signs config is we'd lose the higher-level logic we're using now. Aside from this exception, the logic is clearly expressing "whenever we're heading toward the unique endpoint of a branch", and we'd lose that insight if it was just flags in the config.

PaulJKim: That is true. I guess I'm also wondering at the same time if we need to be concerned with the specific endpoints of a branch at all here. To me it's almost like the endpoints actually don't matter so much as the direction. You could potentially sum up the few versions of the existing function as "only announce the branch when trains are westbound". The only one that doesn't _quite_ fit that is the Green-D Reservoir case, but I think that might just be because Reservoir is a common alternative endpoint of the D branch when trains don't run all the way to Riverside. The direction of trains can be inferred based on what stop we're setting the `announce_branch` flag for, which would mean that we wouldn't need that as a parameter for the function at all. We'd probably want to do some digging to verify whether or not there's a historical reason for announcing branches _only_ for certain direction + endpoint combos though.

PaulJKim: Upon revisiting this, I think I'm actually more in favor of keeping the logic as it is, even if there is a little bit of duplication happening.

The intention behind this PR is essentially "If it's Kenmore, always read out the branch letter". It's a very specific, special case and I think the code expresses that to an appropriate degree. I also don't think any duplication of logic is happening to a degree that is alarming.

Depending on what's decided on w/r/t reading out branch letters all along the Green Line trunk, we may end up being able to just eliminate `route_and_destination_branch_letter`, but I feel like for now this is probably fine for the purposes of treating Kenmore with this special casing. Let me know what you think though!

panentheos (approved):
Looks good.

PR #563, opened 2023-01-20
fix health check test and speed up tests

panentheos:
#### Summary of changes
**Asana Ticket:** [Investigate and resolve intermittent health test errors](https://app.asana.com/0/1201753694073608/1203657802315715/f)

This should fix the intermittent failures on the health check test. The issue is a race condition, where we `send` a message to the module and then immediately check its logs for a response. This adds a short wait before checking, similar to how other tests handle this same situation.

This also reduces some unnecessarily long sleeps in the test code, which results in a ~1 second reduction in total time for `mix test` (3 sec down to 2 sec for me).

Also converts `:timer.sleep` to the more recent `Process.sleep`

### DIFF
--- a/test/engine/alerts_test.exs
+++ b/test/engine/alerts_test.exs
@@ -11,7 +11,7 @@ defmodule Engine.AlertsTest do
           routes_ets_table_name: :routes_ets_table_test
         )
 
-      Process.sleep(500)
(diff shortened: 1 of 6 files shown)


PaulJKim (approved):
LGTM! Thank you for looking into this!

PR #564, opened 2023-01-23
add bus sign type and get bus predictions

panentheos:
#### Summary of changes
**Asana Ticket:** [Create a new "bus" sign type](https://app.asana.com/0/1201753694073608/1203670096087486/f)
**Asana Ticket:** [Get bus predictions](https://app.asana.com/0/1201753694073608/1203670096087488/f)

This adds some initial pieces of the Transitway Engine code base:
* A new "bus" sign type to differentiate the behavior
* A module that downloads bus predictions
* A script that parses bulk data directly from the TE code

NOTE: None of the new modules are enabled by this PR, so this should be a no-op in production.

### DIFF
--- a/lib/engine/bus_predictions.ex
+++ b/lib/engine/bus_predictions.ex
@@ -21,7 +21,7 @@ defmodule Engine.BusPredictions do
     api_key = Application.get_env(:realtime_signs, :api_v3_key)
     http_client = Application.get_env(:realtime_signs, :http_client)
 
-    with {:ok, req} <-
+    with {:ok, %{status_code: 200, body: body}} <-
            http_client.get(
              api_url <> "/predictions",
              if api_key do
@@ -39,17 +39,16 @@ defmodule Engine.BusPredictions do
                "filter[stop]" => Enum.join(Signs.Utilities.SignsConfig.all_bus_stop_ids(), ",")
              }
            ),
-         %{status_code: 200, body: body} <- req,
          {:ok, %{"data" => data, "included" => included}} <- Jason.decode(body) do
       vehicles_lookup =
         for %{
               "type" => "vehicle",
               "id" => id,
               "attributes" => %{"updated_at" => updated_at}
-            } <- included do
-          %{id: id, updated_at: updated_at}
+            } <- included,
+            into: %{} do
+          {id, %{id: id, updated_at: updated_at}}
         end
-        |> index_by(& &1.id)
 
       trips_lookup =
         for %{
@@ -59,10 +58,10 @@ defmodule Engine.BusPredictions do
               "relationships" => %{
                 "route" => %{"data" => %{"id" => route_id}}
               }
-            } <- included do
-          %{id: id, headsign: headsign, route_id: route_id}
+            } <- included,
+            into: %{} do
+          {id, %{id: id, headsign: headsign, route_id: route_id}}
         end
-        |> index_by(& &1.id)
 
       new_predictions =
         for %{
@@ -82,7 +81,8 @@ defmodule Engine.BusPredictions do
             route_id == trips_lookup[trip_id].route_id do
           %{
             direction_id: direction_id,
-            departure_time: Timex.parse!(departure_time, "{ISO:Extended}"),
+            departure_time:
+              if(departure_time, do: Timex.parse!(departure_time, "{ISO:Extended}")),
             route_id: route_id,
             stop_id: stop_id,
             headsign: trips_lookup[trip_id].headsign,
@@ -114,9 +114,5 @@ defmodule Engine.BusPredictions do
 
   defp schedule_update(pid) do
     Process.send_after(pid, :update, 1_000)
-  end
-
-  defp index_by(enum, key_fun) do
-    Enum.into(enum, %{}, &{key_fun.(&1), &1})
   end
 end
--- a/lib/engine/bus_predictions.ex
+++ b/lib/engine/bus_predictions.ex
@@ -1,0 +1,122 @@
+defmodule Engine.BusPredictions do
+  use GenServer
+  require Logger
+
+  def predictions_for_stop(id) do
+    GenServer.call(__MODULE__, {:predictions_for_stop, id})
+  end
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+  end
+
+  def init(_) do
+    schedule_update(self())
+    {:ok, %{predictions: %{}}}
+  end
+
+  def handle_info(:update, state) do
+    schedule_update(self())
+    api_url = Application.get_env(:realtime_signs, :api_v3_url)
+    api_key = Application.get_env(:realtime_signs, :api_v3_key)
+    http_client = Application.get_env(:realtime_signs, :http_client)
+
+    with {:ok, req} <-
+           http_client.get(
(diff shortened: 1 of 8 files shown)


lib/signs/bus.ex
```diff
@@ -0,0 +1,106 @@
...
+  require Logger
+
+  def start_link(sign, opts \\ []) do
+    %{
+      "type" => "bus",
+      "id" => id,
+      "pa_ess_loc" => pa_ess_loc,
+      "text_zone" => text_zone,
+      "audio_zones" => audio_zones,
+      "sources" => sources
```
PaulJKim, line 12: This pattern match does a good job of getting these values and checking for their existence, since the pattern match will throw an exception if any of these are missing in `sign`, but it might be even better to use `Map.fetch!` to get some of these like we do in `realtime.ex`. The advantage is a more specific exception if/when a key is missing from the map.

lib/signs/bus.ex
```diff
@@ -0,0 +1,106 @@
...
+        prediction
+      end
+
+    # Exclude missing headsign and/or display route (error?)
+
+    # Hold prediction in case of count-up
+    # Special case: Hold prediction for inbound SL1 with stale prediction
+
+    _ = predictions
+    _ = config
```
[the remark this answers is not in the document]
panentheos: They're there to suppress the warning about unused values, since they would become errors during the build. That seemed easier than trying to selectively comment everything.

scripts/transitway_engine/parse_config.exs
```diff
@@ -0,0 +1,90 @@
+# This is a temporary script to help port Transitway Engine functionality. It parses the
+# code/config and transforms the data into a format that works for realtime_signs.
+
+# This script currently expects to be run from the root directory of realtime_signs,
+# with a copy of transitway_engine sitting in a sibling directory.
+
+# NOTE: This script will modify/overwrite files, so only run it with a clean working directory.
+
+Mix.install([{:jason, "~> 1.4.0"}])
```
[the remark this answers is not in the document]
panentheos: Ah, thanks for the tip. I agree in general that we should keep dependencies upgraded, but let's do that separately since this script is temporary anyway.

lib/signs/bus.ex
PaulJKim: The match error would look something like `(MatchError) no match of right hand side value: %{a: 1, b: 2, c: 3}` where the right hand side value would be the value of `sign`. You'd have to take a look at the pattern match in the code and compare it with the logged state of `sign` to figure out which key it was that you're missing. Trivial when the map is small, but less so as it gets bigger.

The error from `Map.fetch!` would look something like `(KeyError) key :d not found in: %{a: 1, b: 2, c: 3} (stdlib 3.17) :maps.get(:d, %{a: 1, b: 2, c: 3})`  so we'd immediately know what key was missing from the config.

And yeah, I do agree that the pattern matching is a nice way to convey the structure of the data. But since we'll be including pretty much everything in `sign` in our struct definition, aren't we already pretty clearly conveying the shape of the data?

scripts/transitway_engine/parse_config.exs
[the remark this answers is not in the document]
PaulJKim: True, we won't be reusing this ever once TWE is fully ported so need to sweat it

lib/engine/bus_predictions.ex
```diff
@@ -0,0 +1,122 @@
...
+                "trip" => %{"data" => %{"id" => trip_id}},
+                "vehicle" => %{"data" => vehicle_data}
+              }
+            } <- data,
+            # Multi-route trips will have duplicate predictions for each route.
+            # To filter them out, we only keep the one whose route_id matches the trip's.
+            route_id == trips_lookup[trip_id].route_id do
+          %{
+            direction_id: direction_id,
+            departure_time: Timex.parse!(departure_time, "{ISO:Extended}"),
```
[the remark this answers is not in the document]
PaulJKim: Yeah in Elixir we don't typically catch exceptions in the fashion that is common in other languages. Elixir and Erlang folks adopt a principle they dubbed as "let it crash" where we go ahead and let the process completely fail and depend on supervision trees to automatically restart the crashed process, along with any adjacent processes that are part of the same supervision tree. This is how Elixir/Erlang applications "self-heal" themselves. That approach is more for unexpected errors though, or for handling method calls that fail but we know that we want the process to crash as a result and be restarted. The latter is an instance where we'd want to call the version of a method that ends with a `!`, if it exists.

When we're considering an error that we want to handle and let the larger process live on, we typically will use a `case` statement and pattern match on `{:ok, ...}` on success, or `{:error, ...}` on failure and respond appropriately. Here's a random example of what I'm talking about https://github.com/mbta/realtime_signs/blob/main/lib/content/audio/approaching.ex#L33-L40

lib/signs/bus.ex
```diff
@@ -0,0 +1,111 @@
...
+      # Exclude predictions that are too old
+      |> Enum.reject(&Timex.before?(&1.departure_time, Timex.shift(current_time, seconds: -5)))
+      # Special case: exclude 89.2 OB to Davis
+      |> Enum.reject(&(&1.stop_id == "5104" && String.starts_with?(&1.headsign, "Davis")))
+      # Special case: exclude routes terminating at Braintree (230.4 IB, 236.3 OB)
+      |> Enum.reject(&(&1.stop_id == "38671" && String.starts_with?(&1.headsign, "Braintree")))
+      # Special case: exclude routes terminating at Mattapan, in case those variants of route 24 come back.
+      |> Enum.reject(
+        &(&1.stop_id in ["185", "18511"] && String.starts_with?(&1.headsign, "Mattapan"))
+      )
```
PaulJKim, line 78: We could make this a bit more efficient with the use of [Streams](https://hexdocs.pm/elixir/1.14.3/Stream.html). Each function in the `Enum` module more or less iterates over the entire collectable and creates intermediate lists between each call. Streams are lazy and composable so would save us 4 iterations over the list of predictions in this case. Luckily, their APIs are very similar, if not identical. Check out https://hexdocs.pm/elixir/1.14.3/Stream.html#reject/2!

lib/engine/bus_predictions.ex
```diff
@@ -0,0 +1,118 @@
...
+            stop_id: stop_id,
+            headsign: trips_lookup[trip_id].headsign,
+            updated_at:
+              case vehicle_data do
+                %{"id" => vehicle_id} -> vehicles_lookup[vehicle_id].updated_at
+                _ -> nil
+              end
+          }
+        end
+        |> Enum.group_by(& &1.stop_id)
```
PaulJKim, line 96: I think this is probably fine for now, but probably will be helpful to define a struct for bus predictions later on, along with a corresponding module to help marshal the map data into said struct

lib/signs/bus.ex
panentheos: Done, thanks for the pointer.

lib/engine/bus_predictions.ex
panentheos: Perhaps later on, but since this is still in active development, I'd like to see how the shape of the code evolves before solidifying things too much.

PaulJKim (approved):
LGTM with the understanding that this is still a WIP and won't have any affect in production. Nice start to the porting over of TWE!

PR #569, opened 2023-02-07
support mezzanine mode for bus signs

panentheos:
#### Summary of changes
**Asana Ticket:** [Implement mezzanine mode for bus signs](https://app.asana.com/0/1201753694073608/1203913364994578/f)

Adds mezzanine mode for bus signs. Currently this only affects a single sign (WTC mezzanine). This mode fetches two separate prediction lists and displays them independently. Updates the bus json schema with two new keys, `top_sources` and `bottom_sources`, which are used to configure the sign in this mode.

### DIFF
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -11,6 +11,8 @@ defmodule Signs.Bus do
     :audio_zones,
     :max_minutes,
     :sources,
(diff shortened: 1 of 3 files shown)


PaulJKim (approved):
Makes sense!

PR #570, opened 2023-02-09
handle bus sign overrides

panentheos:
#### Summary of changes
**Asana Ticket:** [Handle config overrides for bus signs](https://app.asana.com/0/1201753694073608/1203913364994582/f)

This makes the bus signs respect the per-sign overrides in the config file. Also implements a special case for the Harvard lower sign.


PaulJKim (approved):
LGTM!

PR #573, opened 2023-02-09
Define headway group one level up

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🐞: Ashmont mezzanine sign second line pages Alewife headways](https://app.asana.com/0/1201753694073608/1203894734388571/f)

This PR involves a fairly small logic change but a larger refactor of the signs config.

By moving `headway_group` up one level in the config, we gain the flexibility to assign each line of the sign to a different headway group. This is **only** relevant for signs that have the ability to page headways on the second line (doubly-sourced signs which are by and large mezzanine signs). The only case where this is applicable is the Ashmont mezzanine sign which should be able to page Mattapan headways on the second line when there are no Mattapan predictions. Right now, it will show Alewife headways.

There are better ways to structure the config, but those would likely require a much larger refactor which will be more time consuming. We can revisit the config and consider refactoring in the future.

Update to Ashmont Mezzanine:
Screenshot 2023-02-10 at 2 12 51 PM


lib/signs/utilities/messages.ex
```diff
@@ -98,13 +98,21 @@ defmodule Signs.Utilities.Messages do
...
-          top_line_source
+          headway_group = Signs.Utilities.SourceConfig.default_headway_group(sign.headway_group)
+          {top_line_source, headway_group}
         else
           %Signs.Realtime{source_config: {_, bottom_line_source}} = sign
-          bottom_line_source
+
+          headway_group =
+            case sign.headway_group do
+              [_, bottom_headway_group] -> bottom_headway_group
```
PaulJKim, line 111: This is giving a "pattern can never match" warning for some reason, but I'm pretty sure it's just dialyzer being wrong 🤷

panentheos: I'm getting this too in my editor, but not when running dialyzer manually. Is this a known issue?

lib/signs/utilities/source_config.ex
```diff
@@ -214,4 +214,8 @@ defmodule Signs.Utilities.SourceConfig do
   def sign_routes({s}) do
     Enum.flat_map(s, &(&1.routes || []))
   end
+
+  def default_headway_group([headway_group, _]), do: headway_group
```
panentheos, line 218: What about completely abstracting access to this field with a helper like `sign_headway_group(sign, line \\ :top)`? Then we could use it in all three places.

panentheos (approved):
Looks good, but see comments.

lib/signs/utilities/messages.ex
PaulJKim: I don't know if it's a documented issue off the top of my head, but I know that ElixirLS can be flaky when it comes to dialyzer.

lib/signs/utilities/source_config.ex
PaulJKim: Good call! This did simplify things a lot and I think we can subsequently reduce the number of places we call the function to just 2. Just pushed the changes!

PR #575, opened 2023-02-14
Announce branch letter at all gl stops

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Update audio announcements ~at Kenmore~ on all of GL](https://app.asana.com/0/1201753694073608/1203674455933121/f)

This PR makes a change so that we _always_ announce the branch letter for Green Line trains. We already do this for westbound trains so this essentially expands the behavior for eastbound trains as well.


panentheos (approved):
Looks good!

PR #577, opened 2023-02-15
handle Chelsea bridge status on bus signs

panentheos:
#### Summary of changes
**Asana Ticket:** [Handle Chelsea bridge logic](https://app.asana.com/0/1201753694073608/1203958349180920/f)

This adds a new engine to get the status of the Chelsea bridge, and displays that information on the relevant signs. The engine is disabled unless you're in "bus mode", and also requires setting a new environment variable with the encoded auth string, which can be found in the password vault.

### DIFF
--- a/lib/engine/chelsea_bridge.ex
+++ b/lib/engine/chelsea_bridge.ex
@@ -1,0 +1,58 @@
+defmodule Engine.ChelseaBridge do
+  use GenServer
+  require Logger
+
+  def bridge_status() do
+    GenServer.call(__MODULE__, {:bridge_status})
+  end
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+  end
+
+  def init(_) do
+    send(self(), :update)
+    {:ok, %{status: %{raised?: nil, estimate: nil}}}
+  end
+
+  def handle_info(:update, state) do
+    Process.send_after(self(), :update, 60_000)
+    url = Application.get_env(:realtime_signs, :chelsea_bridge_url)
+    auth = Application.get_env(:realtime_signs, :chelsea_bridge_auth)
+    http_client = Application.get_env(:realtime_signs, :http_client)
+
+    with {:ok, %{status_code: 200, body: body}} <-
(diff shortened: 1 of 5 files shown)


lib/engine/chelsea_bridge.ex
```diff
@@ -0,0 +1,60 @@
...
+           ),
+         {:ok, data} <- Jason.decode(body) do
+      %{"bridge" => %{"bridgeStatusId" => %{"status" => status}}} = data
+
+      estimate =
+        case data do
+          %{"lift_estimate" => %{"estimate_time" => estimate}} ->
+            String.replace(estimate, ~r/\.\d+$/, "")
+            |> Timex.parse!("{YYYY}-{M}-{D} {h24}:{m}:{s}")
+            |> DateTime.from_naive!("America/New_York")
```
PaulJKim, line 36: These calls will crash the process if they ever fail. Usually if we can anticipate some type of error, we'll handle it in some way and log an error and reserve the Supervisor's automatic restarts for the unexpected stuff, but I could see reason to leave this as-is since I'm not sure if we necessarily want to handle these potential errors and we're expecting a particular format. Food for thought though

panentheos: I've been wrestling with how to handle this and similar code in the other engine. Technically the API could start returning different data at any time, which could cause any number of errors in the downstream handling. Barring some sort of catch-and-log approach, it seems like "letting it crash" is the best way to get early warnings of unexpected changes.

PaulJKim: Yeah that's a fair point. I think you're right that letting is crash does make sense here. Especially given that the engine's rate of polling the API is only once a minute. It'd be something we'd be more concerned about handling gracefully if it was polling more frequently. Reason being that if a process causes the Supervisor to hit the maximum restart frequency (defaults to 3 in 5 seconds) the Supervisor will terminate itself and all it's children and since there's only one Supervisor in Realtime Signs, that would cause a restart of the whole thing

PaulJKim (approved):

PR #579, opened 2023-02-22
construct audio messages for bus signs

panentheos:
#### Summary of changes
**Asana Ticket:** [Create audio messages for bus signs](https://app.asana.com/0/1201753694073608/1204014105313025/f)

This adds the bulk of the logic for generating audio messages for bus signs. Note that the approach here (individual zone audios) is different from what Transitway Engine does (station-wide audios), so there are some behavior differences. There are also a few open questions (marked by TODOs), so this code will likely change further.

This doesn't handle sending out the audio messages; that will be handled in a followup change.

Running this will currently log several errors around missing audio mappings. Some of these will probably need to be uploaded, and some may already be there and just need to be mapped in code.


lib/pa_ess/utilities.ex
```diff
@@ -506,4 +506,170 @@ defmodule PaEss.Utilities do
...
+    "upcoming arrivals" => "550",
+    "is now arriving" => "24055",
+    "upper level departures" => "616",
+    "lower level departures" => "617",
+    "board routes 71 and 73 on upper level" => "618",
+    "departing" => "530",
+    "arriving" => "531",
+    "," => "21012",
+    "minute" => "532",
+    "minutes" => "505"
```
PaulJKim, line 662: Minor: Atoms are generally preferred for static content where it makes sense, particularly as map keys. Mainly because they take up less memory and are faster for comparisons. It doesn't make as much sense for the prior two maps because one is basically all numeric and the other gets the values from the API and we don't want to convert strings to atoms especially from a dynamic source like the API. I also find that atoms are easier to read. Could we opt for atoms for this map? `,` could just be `:comma`

lib/signs/bus.ex
```diff
@@ -156,17 +162,43 @@ defmodule Signs.Bus do
...
+              {@drawbridge_minutes, [PaEss.Utilities.number_var(minutes, :english)]},
+              {@drawbridge_minutes_spanish, [PaEss.Utilities.number_var(minutes, :spanish)]}
+            ]
+        end
+        |> Enum.map(fn {msg, vars} ->
+          %Content.Audio.BusPredictions{message: {:canned, {msg, vars, :audio_visual}}}
+        end)
+      else
+        []
+      end
```
PaulJKim, line 182: I think we discussed this in an earlier PR and agreed to just get things working before breaking stuff out, but thoughts on starting a separate module for audio related things to start? It's getting a little tougher to read/review the code and keep track of existing/new functionality since there's many things going on in `bus.ex` now

PaulJKim (approved):
LGTM! Definitely improves the readability 🎉

PR #580, opened 2023-02-22
Optimize message latency report job

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Optimize Message latency job](https://app.asana.com/0/1201753694073608/1203923825770741/f)

This PR makes some memory optimizations to the message latency report job. Initially, I thought we might be able to batch process the CSVs but it turns out that it's primarily just one CSV that is causing the memory bottleneck which is the `all.csv` file that the script generates by combining all of the CSVs. Given the way percentiles work, we need all of the rows in memory so that we can sort them and calculate where the percentiles land and since `Enum.sort`'s space complexity is O(n), this single file uses the same amount of memory at one time as all of the others combined throughout the course of a run of the job. Therefore, batching likely won't be of much help here.

I did make a change to find the 95th and 99th percentiles in a single pass rather than two to help time efficiency a bit. But to help space complexity, I simply removed a number of fields that we don't need from the data structure that the script works with which slimmed things down quite a bit. The script now uses about 1/3-1/4 of the memory it was using before which is still a good improvement.

I also added a tag to the logs to make it easier to query logs related to this job as well as logging around the total size of the data we get to ARINC which can help us monitor how much data we're actually getting from them without having to manually check S3.

### DIFF
--- a/lib/jobs/message_latency_report.ex
+++ b/lib/jobs/message_latency_report.ex
@@ -94,6 +94,9 @@ defmodule Jobs.MessageLatencyReport do
       |> Stream.filter(&filter_by_date(&1, date))
       |> Stream.map(fn row -> row[:seconds] end)
       |> Enum.sort()
+      |> tap(fn _ ->
+        Logger.info("message_latency_report: Memory in use after sort: #{:erlang.memory(:total)}")
+      end)
       |> Stream.with_index()
       |> get_percentile_rows()
 
--- a/lib/jobs/message_latency_report.ex
+++ b/lib/jobs/message_latency_report.ex
@@ -38,6 +38,7 @@ defmodule Jobs.MessageLatencyReport do
   defp analyze_files_for_date(date) do
     with {:ok, response} <- get_zip_file(date),
          {:ok, files} <- :zip.unzip(response.body) do
+      record_log_files_size(files)
       unzipped_directory = String.replace(date, "-", "")
 
       case :os.type() do
@@ -51,10 +52,7 @@ defmodule Jobs.MessageLatencyReport do
           :os.cmd(:"type #{unzipped_directory}\\*.csv > #{unzipped_directory}\\all.csv")
       end
 
-      [
-        get_csv_row("#{unzipped_directory}/all.csv", date)
-        | Enum.map(files, &get_csv_row(&1, date))
-      ]
+      Enum.map(["#{unzipped_directory}/all.csv" | files], &get_csv_row(&1, date))
       |> format_csv_data()
       |> put_csv_in_s3(date)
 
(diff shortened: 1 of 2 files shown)


lib/jobs/message_latency_report.ex
```diff
@@ -30,14 +30,15 @@ defmodule Jobs.MessageLatencyReport do
       ) do
     Enum.each(0..(days_to_analyze - 1), fn diff ->
       date = start_date |> Date.add(-diff) |> Date.to_string()
-      Logger.info("Generating message latency report for #{date}...")
+      Logger.info("message_latency_report: Generating message latency report for #{date}...")
```
panentheos, line 33: If we want this prefix everywhere, perhaps we should create a helper function that applies it. Also there's one `Logger.info` in `format_csv_data` that didn't get the prefix.

lib/jobs/message_latency_report.ex
```diff
@@ -210,13 +198,27 @@ defmodule Jobs.MessageLatencyReport do
...
-      Enum.find(rows_with_index, {[], nil}, fn {_row, index} ->
-        index == percentile_index
-      end)
+  defp get_percentiles(rows_with_index) do
+    num_rows = Enum.count(rows_with_index)
+    percentile_95_index = calculate_percentile_index(num_rows, @percentile_95)
+    percentile_99_index = calculate_percentile_index(num_rows, @percentile_99)
 
-    percentile_row
+    {Enum.reduce_while(rows_with_index, {[], []}, fn {current_row, index},
```
panentheos, line 206: I think `nil` would be a better guard value than `[]`, since these don't end up as lists, right?

lib/jobs/message_latency_report.ex
```diff
@@ -210,13 +198,27 @@ defmodule Jobs.MessageLatencyReport do
...
+                                                     {percentile_95, percentile_99} ->
+       cond do
+         percentile_95 != [] and percentile_99 != [] ->
```
panentheos, line 209: nit: Since we know we're done at 99, could we halt there instead of doing one more loop?

lib/jobs/message_latency_report.ex
```diff
@@ -210,13 +198,27 @@ defmodule Jobs.MessageLatencyReport do
...
+           {:halt, {percentile_95, percentile_99}}
+
+         index == percentile_95_index ->
+           {:cont, {current_row, percentile_99}}
+
+         index == percentile_99_index ->
```
panentheos, line 215: This may never be an issue, but if the data set is small enough, the 95th and 99th percentile could be the same, in which case the `cond` would end up skipping the 99th. If we wanted to fix that we'd need to e.g. pipe the accumulator through both checks.

lib/jobs/message_latency_report.ex
```diff
@@ -225,6 +227,20 @@ defmodule Jobs.MessageLatencyReport do
     |> round()
   end
 
+  defp record_log_files_size(log_files) do
+    Enum.reduce(log_files, 0, fn file, acc ->
```
panentheos, line 231: nit: We could `map` and `sum` instead of `reduce` here, if desired

panentheos (approved):
See comments, but the overall approach looks good!

lib/jobs/message_latency_report.ex
PaulJKim: Oh great catch, yeah they used to return as lists before the refactor. The actual values should be floats though so I think I'll just opt to use two zeros

lib/jobs/message_latency_report.ex
PaulJKim: good point!

lib/jobs/message_latency_report.ex
PaulJKim: Hmm yeah good point. It's certainly possible considering the amount of data that ARINC as sent us varies wildly from past experience. I'll make a change to make sure we don't skip the 99th

lib/jobs/message_latency_report.ex
PaulJKim: Will probably just keep it as a reduce since the difference is minimal in this case

PR #581, opened 2023-02-24
Revisit jfk umass platform announcements

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Revisit JFK/UMass platform prediction audio](https://app.asana.com/0/1201753694073608/1203683812842464/f)

This adds a tweak to the logic around JFK/UMass platform prediction audio announcements. When the train more than 5 but less than 10 minutes away, we will announce that the platform "will be announced soon". When the train is 10 minutes or more away, we will announce that the platform "will be announced when the train is closer".


panentheos (approved):
Looks good!

PR #585, opened 2023-03-01
send scheduled bus audio messages

panentheos:
#### Summary of changes
**Asana Ticket:** [Send bus audio messages](https://app.asana.com/0/1201753694073608/1204067334255204/f)

This closes the loop and sends out the computed bus audio messages on the specified schedule. Also corrects the predictions preamble for south station, and skips it if there are no predictions.

### DIFF
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -19,11 +19,14 @@ defmodule Signs.Bus do
     :top_sources,
     :bottom_sources,
     :chelsea_bridge,
+    :read_loop_interval,
+    :read_loop_offset,
     :config_engine,
     :prediction_engine,
     :prev_predictions,
     :current_messages,
-    :last_update
+    :last_update,
+    :last_read_time
   ]
   defstruct @enforce_keys
 
@@ -38,11 +41,14 @@ defmodule Signs.Bus do
       top_sources: parse_sources(sign["top_sources"]),
       bottom_sources: parse_sources(sign["bottom_sources"]),
       chelsea_bridge: sign["chelsea_bridge"],
+      read_loop_interval: Map.fetch!(sign, "read_loop_interval"),
+      read_loop_offset: Map.fetch!(sign, "read_loop_offset"),
       config_engine: opts[:config_engine] || Engine.Config,
       prediction_engine: opts[:prediction_engine] || Engine.BusPredictions,
       prev_predictions: [],
       current_messages: {nil, nil},
-      last_update: nil
+      last_update: nil,
+      last_read_time: Timex.now()
     }
 
     GenServer.start_link(__MODULE__, state)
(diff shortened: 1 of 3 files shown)


lib/signs/bus.ex
```diff
@@ -318,6 +329,17 @@ defmodule Signs.Bus do
          ))
   end
 
+  defp should_read(current_time, state) do
```
PaulJKim, line 332: Minor, but by Elixir naming conventions we should add a `?` to the end of this function to indicate that it returns a boolean

lib/signs/bus.ex
```diff
@@ -135,23 +142,29 @@ defmodule Signs.Bus do
...
       if should_update({top, bottom}, current_time, state) do
         MessageQueue.update_sign({pa_ess_loc, text_zone}, top, bottom, 180, :now)
         %{state | current_messages: {top, bottom}, last_update: current_time}
       else
         state
       end
+    end)
+    |> then(fn state ->
+      if should_read(current_time, state) do
+        if audios != [], do: MessageQueue.send_audio({pa_ess_loc, audio_zones}, audios, 5, 60)
```
PaulJKim, line 156: Question: is the value of the timeout taken from TWE? Since I know there is potential for a lot more audio readouts for bus compared to rail, I'm wondering if there's higher chance for audio to timeout. Or do you think each message should be sufficiently below the timeout considering we iterate through the list and process each entry in `audios` separately?

lib/signs/bus.ex
```diff
@@ -456,6 +478,20 @@ defmodule Signs.Bus do
     ]
   end
 
+  defp add_preamble([], _), do: []
+
+  # Special case: South Station is currently the only terminal we handle, so it should say
+  # "departures" instead of "arrivals".
```
PaulJKim, line 484: Hmm interesting.. is this actually true? I think we also have PA/ESS at the Forest Hills busway and Nubian, at least. Or are those not necessarily terminals for some reason? 🤔

panentheos: I rechecked, TWE only says "departures" at South Station (and also Harvard, but that's a special case I haven't looked into fully yet). But I don't know if that's the behavior we want going forward, so I'll check.

lib/signs/bus.ex
```diff
@@ -456,6 +478,20 @@ defmodule Signs.Bus do
...
+  defp add_preamble(items, state) do
+    [
+      case state.id do
+        "Silver_Line.South_Station_EB" -> [:upcoming_departures]
+        _ -> [:upcoming_arrivals]
```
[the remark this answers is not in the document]
panentheos: This depends on the answer to the above question. If there are multiple places we want to do this, then a config setting makes sense. If it's just the one, then I think the special override is cleaner.

lib/signs/bus.ex
panentheos: Good catch. TWE uses a timeout of 180, so I'll use that instead.

lib/signs/bus.ex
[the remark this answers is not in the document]
PaulJKim: Yeah, I agree on that

PaulJKim (approved):
Overall LGTM! Up to you what happens to the preamble logic after you check on if there are potentially more places we'd want to say "departures at" since we're agreed on how to decide which direction to take.

PR #587, opened 2023-03-06
Add check before displaying paging headways

PaulJKim:
#### Summary of changes
**Asana Ticket:** [New paging headways logic doesn't check against schedules data](https://app.asana.com/0/1201753694073608/1204118945498884/f)

This PR adds a check by calling `display_headways?` before deciding to show paging headways. The check uses schedules data to determine if we're within scheduled service hours and mirrors what we do for normal headway messages.

Tested by faking the current time as after service hours in the code itself. Result:
Screenshot 2023-03-06 at 11 36 15 AM

### DIFF
--- a/lib/signs/utilities/headways.ex
+++ b/lib/signs/utilities/headways.ex
@@ -36,7 +36,13 @@ defmodule Signs.Utilities.Headways do
         %Content.Message.Empty{}
 
       headways ->
-        config |> source_list_destination() |> get_paging_headway_message(headways)
+        stop_ids = get_stop_ids(config)
+
+        if display_headways?(sign, stop_ids, current_time, headways) do
+          config |> source_list_destination() |> get_paging_headway_message(headways)
+        else
+          %Content.Message.Empty{}
+        end
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/signs/utilities/headways.ex
```diff
@@ -56,6 +62,9 @@ defmodule Signs.Utilities.Headways do
   defp get_stop_ids(sign, nil), do: Signs.Utilities.SourceConfig.sign_stop_ids(sign.source_config)
   defp get_stop_ids(sign, config), do: [sign.headway_stop_id || config.stop_id]
 
+  @spec get_stop_ids(list(SourceConfig.source())) :: [String.t()]
+  defp get_stop_ids(sources), do: Enum.map(sources, & &1.stop_id)
```
panentheos, line 66: nit: Since this is tiny and single-use, perhaps we should inline it.

panentheos (approved):
Looks good, see one minor comment.

lib/signs/utilities/headways.ex
PaulJKim: Fair point! Will make that tweak

PR #588, opened 2023-03-14
refactor config engine test

panentheos:
#### Summary of changes

This reworks the config engine test to abstract away knowledge of the underlying ETS tables. The goal is to free the tests from having to explicitly create tables during setup. This should reduce spontaneous failures when the internal implementation of the engine changes.


PaulJKim (approved):
LGTM

PR #589, opened 2023-03-14
implement misc bus features

panentheos:
#### Summary of changes

This includes several smaller features:
* "Debounces" prediction times, to ensure that stale predictions are never revived later due to changes around the zero mark (see `prediction_stale?`)
* Suppresses all SL route names, as well as route names at Nubian A (per decision)
* Loads the Chelsea bridge flag from the config and uses it to control the bridge messages (see `bridge_enabled?`)
* Implements combined audio readout for both sides of Nubian E (see `extra_audio_sources`)
* Implements immediate bridge announcement at waterfront stops when the status changes to raised (see `should_announce_drawbridge?`)
* Fixes a math bug in the bridge status calculation

### DIFF
--- a/lib/engine/config.ex
+++ b/lib/engine/config.ex
@@ -13,6 +13,7 @@ defmodule Engine.Config do
   @type state :: %{
           table_name_signs: term(),
           table_name_headways: term(),
+          table_name_chelsea_bridge: term(),
           current_version: version_id,
           time_fetcher: (() -> DateTime.t())
         }
@@ -21,6 +22,7 @@ defmodule Engine.Config do
 
   @table_signs :config_engine_signs
   @table_headways :config_engine_headways
+  @table_chelsea_bridge :config_engine_chelsea_bridge
 
   def start_link([]) do
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
@@ -40,6 +42,14 @@ defmodule Engine.Config do
     Headways.get_headway(table_name, {headway_group, time_period})
   end
 
+  @spec chelsea_bridge_config(:ets.tab()) :: :off | :auto
+  def chelsea_bridge_config(table_name \\ @table_chelsea_bridge) do
+    case :ets.lookup(table_name, :status) do
+      [{_, "off"}] -> :off
+      _ -> :auto
+    end
+  end
+
   def update(pid \\ __MODULE__) do
     schedule_update(pid, 0)
   end
@@ -49,6 +59,7 @@ defmodule Engine.Config do
     state = %{
       table_name_signs: @table_signs,
       table_name_headways: @table_headways,
+      table_name_chelsea_bridge: @table_chelsea_bridge,
       current_version: nil,
       time_fetcher: opts[:time_fetcher] || fn -> DateTime.utc_now() end
     }
@@ -63,6 +74,13 @@ defmodule Engine.Config do
   def create_tables(state) do
     :ets.new(state.table_name_signs, [:set, :protected, :named_table, read_concurrency: true])
     Headways.create_table(state.table_name_headways)
+
+    :ets.new(state.table_name_chelsea_bridge, [
+      :set,
+      :protected,
+      :named_table,
+      read_concurrency: true
+    ])
   end
 
   @spec handle_info(:update, state) :: {:noreply, state}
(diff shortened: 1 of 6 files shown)


lib/signs/bus.ex
```diff
@@ -177,96 +202,123 @@ defmodule Signs.Bus do
...
+      prev = prev_predictions_lookup[prediction_key(prediction)]
+
+      # Prediction times can sometimes jump around from one update to the next, so we adjust them
+      # in certain cases. If the new prediction would count up by 1 minute, or would revive a
+      # previously stale entry, just keep the old time.
+      departure_time =
+        if prev &&
+             (prediction_minutes(prediction, current_time) ==
+                prediction_minutes(prev, current_time) + 1 ||
+                prediction_stale?(prev, current_time)) do
```
PaulJKim, line 220: I'm a little confused on why we want to keep the previously stale entry. If it's stale and we have a fresh prediction, wouldn't we always prefer that one? Or is it so that the stale prediction eventually gets filtered out and we never display it?

lib/signs/bus.ex
```diff
@@ -340,6 +393,23 @@ defmodule Signs.Bus do
     period.(current_time) != period.(last_read_time)
   end
 
+  # In addition to periodic audio messages, we also read the drawbridge message by itself if all
+  # of the following are true:
+  # 1. the drawbridge just went up
+  # 2. drawbridge messages are enabled
+  # 3. we are at a stop that is impacted, but does not show visual drawbridge messages
```
PaulJKim, line 400: Clarifying question: how is this 3rd condition expressed in the logic below?

lib/engine/config.ex
```diff
@@ -63,6 +74,13 @@ defmodule Engine.Config do
   def create_tables(state) do
     :ets.new(state.table_name_signs, [:set, :protected, :named_table, read_concurrency: true])
     Headways.create_table(state.table_name_headways)
+
+    :ets.new(state.table_name_chelsea_bridge, [
```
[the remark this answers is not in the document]
panentheos: Yeah, just consistency. Though I also wonder: Is the overhead of process calls substantial enough that we'd look to avoid it at the scale we're at, and that's why we seem to prefer ETS?

lib/signs/bus.ex
panentheos: Watching the production data, I sometimes see cases where a prediction (seconds) goes e.g. `..., -1, -2, -3, -4, -5, -6, 4, ...`. Without this logic, the prediction drops off the sign momentarily, then comes back when the number jumps back up. With the logic, once it crosses the threshold and becomes stale, we don't allow it to reappear.

I suspect we don't see this in practice today, because TWE updates much less often (> 8 seconds), and the bouncing doesn't seem to last that long.

lib/signs/bus.ex
panentheos: It's `chelsea_bridge == "audio"`. When `chelsea_bridge == "audio_visual"`, we append a visual message to the sign (and don't trigger this immediate readout). There's a TWE comment that explains the rationale, I'll add that for clarity.

lib/signs/bus.ex
PaulJKim: Ah interesting. So that means that a stale prediction never has a chance of being showed again then, even if the prediction gets "reset" or adjusted to a proper value (don't know for sure if that actually happens in real life)? I imagine this is why some riders experience waiting for a bus while watching the countdown go all the way down and disappear, only for no bus to arrive

panentheos: Right. As long as the prediction is still positive, we'll show adjusted numbers as we get them (except for preventing count-up-by-1). Once it becomes stale, though, we're preventing it from reappearing.

From what I've seen, there are two issues:
1. Bad predictions. Probably a multitude of interacting causes, but I've observed prediction deltas count all the way to -300 seconds or so before disappearing. Presumably that corresponds to a bus just not showing up when it's supposed to. We made a decision in TWE to not show anything < -5 seconds, and that got ported over.
2. Predictions "bouncing" by small amounts, especially around the zero mark. That coupled with the staleness threshold was causing previously removed predictions to reappear. The logic here addresses this problem specifically.

PaulJKim: Gotcha, thanks for the explainer! Yeah I guess this is probably mostly a symptom of bad bus predictions. Good to know how we're handling it though!

lib/pa_ess/utilities.ex
```diff
@@ -646,6 +647,8 @@ defmodule PaEss.Utilities do
 
   @atom_take_lookup %{
     the_next_bus_to: "543",
+    the_next: "501",
+    the_following: "667",
```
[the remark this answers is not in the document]
panentheos: It hasn't been uploaded yet. I'll come through and update it after it's there.

PR #592, opened 2023-03-17
remove plug_cowboy

panentheos:
#### Summary of changes

This dependency isn't being used any more, so let's remove it. Also cleans up some other unused entries in the lockfile.


PaulJKim:
I thought cowboy is needed by Phoenix 🤔 . It doesn't run by default unless you start up `RealtimeSignsWeb.Endpoint` which needs value for `MONITOR_SIGN_SCU_PORT` to be set as an env var (that will change with my automatic failover PR though).

PR #593, opened 2023-03-17
Aggregate audio across zones

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate the ability to aggregate audio messages and play in succession](https://app.asana.com/0/1201753694073608/1203549328375594/f)

POC for aggregating audio across audio zones. It ends up skipping sending messages that don't use the "pure take id" format to the headend for simplicity. I think it's a relatively small subset of messages that don't do this, but large enough to where it would be a fair amount of work to convert all of them which isn't important at the moment.

This POC for now skips alert messages because I think those would require some more product discussion to determine how to handle. I stuck to the more straightforward ones that I felt I could come up with reasonable business logic about on my own. Currently, the logic prioritizes audio in the following order:

1. Top line stops away messages
2. Boarding predictions
3. Arriving predictions
4. Approaching predictions
5. Normal predictions
6. Bottom line stops away messages
7. Paging headways (these are always on a bottom line)
8. Normal, double-line headways

### DIFF
--- a/lib/pa_ess/http_updater.ex
+++ b/lib/pa_ess/http_updater.ex
@@ -165,8 +165,10 @@ defmodule PaEss.HttpUpdater do
   def process({:send_audio, [{station, zones}, audios, priority, timeout]}, state) do
     case audios do
       list when is_list(list) ->
-        for audio <- list do
-          process_send_audio(station, zones, audio, priority, timeout, state)
+        combined_audios = combine_audio(list) |> IO.inspect()
+
+        for audio <- combined_audios do
+          process_send_combined_audio(station, zones, audio, priority, timeout, state)
         end
 
       {a1, a2} ->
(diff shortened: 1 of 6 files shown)


lib/signs/utilities/reader.ex
```diff
@@ -35,15 +35,84 @@ defmodule Signs.Utilities.Reader do
...
+        for {_, {_, bottom_content} = bottom} <- all_content,
+            is_struct(
+              bottom_content,
+              Content.Message.Headways.Paging
+            ) do
+          bottom
+        end
+
+      sorted_content =
+        top_stopped_trains ++ predictions ++ bottom_stopped_trains ++ paging_headways ++ headways
```
panentheos, line 101: I like how this is structured, taking all the available content and imposing a high-level strategy on how it's presented.

lib/signs/utilities/reader.ex
```diff
@@ -35,15 +35,84 @@ defmodule Signs.Utilities.Reader do
...
+
+      Signs.Utilities.Audio.from_content_list(sorted_content)
+      |> then(fn audios -> send_audios(sign, audios) end)
+
+      {true, sign}
+    else
+      case Signs.Utilities.Audio.from_sign(sign) do
```
panentheos, line 108: Hopefully we could fold this logic into the above, and come up with something general that works for all cases.

lib/signs/realtime.ex
```diff
@@ -173,6 +187,18 @@ defmodule Signs.Realtime do
     end)
   end
 
+  defp add_extra_audio(%{extra_audio_sign_ids: extra_audio_sign_ids} = sign) do
+    Enum.map(extra_audio_sign_ids, fn sign_id ->
+      Signs.Realtime.get_sign_state(sign_id)
+      |> get_current_contents()
+    end)
+    |> then(fn content -> %Signs.Realtime{sign | extra_audio_content: content} end)
```
panentheos, line 195: Ideally we wouldn't store this in state, since it's only needed for the current iteration, and keeping state minimal should be a goal.

lib/pa_ess/http_updater.ex
```diff
@@ -165,8 +165,10 @@ defmodule PaEss.HttpUpdater do
   def process({:send_audio, [{station, zones}, audios, priority, timeout]}, state) do
     case audios do
       list when is_list(list) ->
-        for audio <- list do
-          process_send_audio(station, zones, audio, priority, timeout, state)
+        combined_audios = combine_audio(list)
```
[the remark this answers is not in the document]
PaulJKim: Ah, didn't realize that the list version was used by the bus code. Agreed that it should happen further up the stack. It should certainly be doable, but there's some friction since it's a departure from Realtime Signs' typical pattern where the structs get converted to message/take ids at basically the last possible second.

Certainly something to consider pursuing though if/when we come back to this task

lib/signs/realtime.ex
```diff
@@ -101,17 +105,22 @@ defmodule Signs.Realtime do
...
-    GenServer.start_link(__MODULE__, sign)
+    GenServer.start_link(__MODULE__, sign, name: String.to_atom(sign.id))
   end
 
   def init(sign) do
     schedule_run_loop(self())
     {:ok, sign}
   end
 
+  def get_sign_state(sign_id) do
```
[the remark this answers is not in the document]
PaulJKim: Yeah, it does add a bit of complexity but I think it's still very easy to reason about the interactions since Elixir is optimized for this sort of thing. And because the signs are just sharing their current state which is being updated once a second anyway, I think doing it this way actually isn't far off from refetching the predictions and doing the same calculations from the source data and is a bit more efficient if anything

PR #594, opened 2023-03-20
Automate cutover to backup headend server

PaulJKim:
#### Summary of changes
**Asana Ticket:** [[#1558] Improve failover process for Realtime Signs](https://app.asana.com/0/1189528452014899/1202213418523535/f)

This PR is adding two main features
1. A URL that ARINC can utilize to point Realtime Signs at a different headend server
2. Logic to the config engine that polls the value of the active headend server from S3

The config value will only be updated in S3 if it is valid and different from what is was before.

The value will only be read into the app config if the application is just starting up, or if there have been at least two reads from S3 with the same value. The latter is a safeguard against potential (albeit unlikely) false positives as suggested by Andrew Daniels from ARINC.

A few things will be needed as follow-up steps
1. Set up ahead of time a `headend.json` file for realtime signs prod containing the required value 
2. Update S3 permissions in Terraform
3. ARINC starts their script that will ping us consistently with the currently active host's ip
4. Remove SIGN_HEAD_END_HOST from secrets

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -7,7 +7,6 @@ config :realtime_signs, RealtimeSignsWeb.Endpoint,
 
 if config_env() != :test do
   config :realtime_signs,
-    sign_head_end_host: System.get_env("SIGN_HEAD_END_HOST"),
     sign_ui_url: System.get_env("SIGN_UI_URL"),
     sign_ui_api_key: System.get_env("SIGN_UI_API_KEY"),
     trip_update_url: System.get_env("TRIP_UPDATE_URL", "https://s3.amazonaws.com/mbta-gtfs-s3/rtr/TripUpdates_enhanced.json"),
@@ -23,7 +22,8 @@ if config_env() != :test do
     message_log_zip_url: System.get_env("MESSAGE_LOG_ZIP_URL"),
     message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
-    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER")
+    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
+    s3_active_headend_path: System.get_env("HEADEND_S3_PATH")
 end
 
 message_log_job =
--- a/lib/engine/config.ex
+++ b/lib/engine/config.ex
@@ -15,7 +15,8 @@ defmodule Engine.Config do
           table_name_headways: term(),
           table_name_chelsea_bridge: term(),
           current_version: version_id,
-          time_fetcher: (() -> DateTime.t())
+          time_fetcher: (() -> DateTime.t()),
+          last_active_headend_ip: String.t()
         }
 
   @type sign_config :: :auto | :headway | :off | {:static_text, {String.t(), String.t()}}
@@ -61,10 +62,12 @@ defmodule Engine.Config do
       table_name_headways: @table_headways,
       table_name_chelsea_bridge: @table_chelsea_bridge,
       current_version: nil,
-      time_fetcher: opts[:time_fetcher] || fn -> DateTime.utc_now() end
+      time_fetcher: opts[:time_fetcher] || fn -> DateTime.utc_now() end,
+      last_active_headend_ip: nil
     }
 
     schedule_update(self())
+    send(self(), :update_active_headend)
 
     create_tables(state)
 
@@ -118,15 +121,30 @@ defmodule Engine.Config do
           state.current_version
       end
 
+    {:noreply, Map.put(state, :current_version, latest_version)}
+  end
+
+  def handle_info(:update_active_headend, state) do
+    schedule_update_active_headend(self())
+    updater = Application.get_env(:realtime_signs, :external_config_getter)
+
     case updater.get_active_headend_ip() do
       {:ok, active_headend_ip} ->
-        Application.put_env(:realtime_signs, :sign_head_end_host, active_headend_ip)
+        # Update active headend ip if startup or after at least two consecutive checks
+        if state.last_active_headend_ip == nil or
+             active_headend_ip == state.last_active_headend_ip do
+          Application.put_env(:realtime_signs, :sign_head_end_host, active_headend_ip)
+          Logger.info("active_headend_ip: current: #{active_headend_ip}")
+        else
+          Logger.info("active_headend_ip: pending update to: #{active_headend_ip}")
+        end
+
+        {:noreply, Map.put(state, :last_active_headend_ip, active_headend_ip)}
 
       _ ->
         Logger.warn("active_headend_ip: was not able to fetch active headend ip")
+        {:noreply, state}
     end
-
-    {:noreply, Map.put(state, :current_version, latest_version)}
   end
 
   def handle_info(msg, state) do
@@ -183,5 +201,9 @@ defmodule Engine.Config do
 
   defp schedule_update(pid, time \\ 1_000) do
     Process.send_after(pid, :update, time)
+  end
+
+  defp schedule_update_active_headend(pid, time \\ 10_000) do
+    Process.send_after(pid, :update_active_headend, time)
   end
 end
--- a/lib/engine/config.ex
+++ b/lib/engine/config.ex
@@ -118,6 +118,14 @@ defmodule Engine.Config do
           state.current_version
       end
 
+    case updater.get_active_headend_ip() do
+      {:ok, active_headend_ip} ->
+        Application.put_env(:realtime_signs, :sign_head_end_host, active_headend_ip)
+
+      _ ->
+        Logger.warn("active_headend_ip: was not able to fetch active headend ip")
+    end
+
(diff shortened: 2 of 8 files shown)


lib/engine/config.ex
```diff
@@ -121,6 +124,29 @@ defmodule Engine.Config do
     {:noreply, Map.put(state, :current_version, latest_version)}
   end
 
+  def handle_info(:update_active_headend, state) do
+    schedule_update_active_headend(self())
+    updater = Application.get_env(:realtime_signs, :external_config_getter)
+
+    case updater.get_active_headend_ip() do
+      {:ok, active_headend_ip} ->
+        # Update active headend ip if startup or after at least two consecutive checks
```
panentheos, line 133: I don't understand why we need to wait for two consecutive results. Can you clarify?

config/runtime.exs
```diff
@@ -23,7 +22,8 @@ if config_env() != :test do
     message_log_zip_url: System.get_env("MESSAGE_LOG_ZIP_URL"),
     message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
-    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER")
+    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
+    s3_active_headend_path: System.get_env("HEADEND_S3_PATH")
```
panentheos, line 26: Does this need to be parameterized? I see that the other paths are parameterized, but assuming the bucket itself is already parameterized per-environment, could we decide to use the same naming scheme for the files, and simplify the configuration?

lib/external_config/local.ex
```diff
@@ -16,4 +16,12 @@ defmodule ExternalConfig.Local do
       {etag, Jason.decode!(file)}
     end
   end
+
+  def get_active_headend_ip() do
+    nil
+  end
+
+  def put_active_headend_ip(_ip) do
+    nil
```
panentheos, line 25: This should return something like `{:ok, nil}` so it matches up with the real implementation.

lib/engine/config.ex
```diff
@@ -176,4 +202,8 @@ defmodule Engine.Config do
   defp schedule_update(pid, time \\ 1_000) do
     Process.send_after(pid, :update, time)
   end
+
+  defp schedule_update_active_headend(pid, time \\ 10_000) do
+    Process.send_after(pid, :update_active_headend, time)
```
panentheos, line 207: Do we need to continuously read this? If we expect all writes to be done through this app, then wouldn't it be sufficient to just read it once on startup?

lib/external_config/s3.ex
```diff
@@ -30,4 +30,40 @@ defmodule ExternalConfig.S3 do
...
+    s3_client = Application.get_env(:realtime_signs, :s3_client)
+    aws_client = Application.get_env(:realtime_signs, :aws_client)
+    bucket = Application.get_env(:realtime_signs, :s3_bucket)
+    path = Application.get_env(:realtime_signs, :s3_active_headend_path)
+
+    case s3_client.put_object(bucket, path, Jason.encode!(%{active_headend_ip: ip}))
+         |> aws_client.request() do
+      {:ok, response} ->
+        Logger.info("active_headend_ip: config changed to: #{ip}")
+        {:ok, response}
```
panentheos, line 62: If we don't set the config value here, this won't take effect for 10 to 20 seconds until the next scheduled reads happen.

lib/engine/config.ex
PaulJKim: It's a suggestion by Andrew to account for potential false positives. He thinks they're unlikely but he suggested waiting for at least two in a row to be safe.

I'm not sure if it's really necessary and am fine with removing it. But it also doesn't really hurt to have the check imo. It does create some potential for delay in the cutover but we're able to handle that and it would be basically imperceptible from the rider perspective

config/runtime.exs
PaulJKim: hmm yeah that's true, we could probably just hardcode it in and not even have it as a config value honestly

lib/external_config/s3.ex
PaulJKim: Right, but we need the delay there if we want to have the "two consecutive" check in play. If we end up removing that, then we would update the config right away here.

config/runtime.exs
PaulJKim: Went with just hardcoding it into `runtime.exs` but let me know if you think there's a better spot for it. Mainly just did that to minimize changes to the PR

config/runtime.exs
```diff
@@ -23,7 +22,8 @@ if config_env() != :test do
     message_log_zip_url: System.get_env("MESSAGE_LOG_ZIP_URL"),
     message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
-    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER")
+    message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
+    s3_active_headend_path: "/headend.json"
```
panentheos, line 26: Now that this doesn't change at runtime, how about just putting it into a module constant in `ExternalConfig.S3`?

lib/external_config/s3.ex
```diff
@@ -30,4 +30,40 @@ defmodule ExternalConfig.S3 do
...
+    path = Application.get_env(:realtime_signs, :s3_active_headend_path)
+
+    case s3_client.get_object(bucket, path)
+         |> aws_client.request() do
+      {:ok, response} ->
+        body = Jason.decode!(response.body)
+        {:ok, body["active_headend_ip"]}
+
+      {:error, e} ->
+        Logger.error("active_headend_ip: s3 response error: #{inspect(e)}")
```
panentheos, line 47: nit: Since we're already logging errors at the call site, we could just return the error here and have that one print it out.

lib/monitoring/headend.ex
```diff
@@ -0,0 +1,24 @@
...
+    valid? = ip == "172.20.145.20" or ip == "172.20.145.22"
+    change? = ip != Application.get_env(:realtime_signs, :sign_head_end_host)
+
+    cond do
+      not valid? ->
+        Logger.warn("active_headend_ip: invalid ip provided: #{ip}")
+        {:bad_request, "bad request"}
+
+      change? ->
+        with {:ok, _} = response <- ExternalConfig.S3.put_active_headend_ip(ip) do
```
panentheos, line 14: This should read the `external_config_getter` config value rather than assuming `ExternalConfig.S3`

lib/realtime_signs_web/controllers/monitoring_controller.ex
```diff
@@ -39,4 +40,17 @@ defmodule RealtimeSignsWeb.MonitoringController do
 
     send_resp(conn, 200, "")
   end
+
+  def update_active_headend(conn, %{"ip" => active_ip} = _params) do
```
panentheos, line 44: Are there any security considerations with exposing this endpoint? I assume we're primarily relying on the network topology to restrict access, but since this impacts the running system, I wanted to check.

lib/realtime_signs_web/controllers/monitoring_controller.ex
```diff
@@ -39,4 +40,17 @@ defmodule RealtimeSignsWeb.MonitoringController do
...
+    case Headend.update_active_headend_ip(active_ip) do
+      {:ok, _} ->
+        send_resp(conn, 200, "ok")
+
+      {:bad_request, message} ->
+        send_resp(conn, 400, message)
+
+      {:error, _} ->
+        send_resp(conn, 500, "Internal server error")
```
panentheos, line 53: It probably doesn't matter on way or another, but do we know if ARINC is planning on doing anything with the returned status code?

config/runtime.exs
PaulJKim: Sure thing!

lib/monitoring/headend.ex
PaulJKim: ahh great catch, thank you!

lib/realtime_signs_web/controllers/monitoring_controller.ex
PaulJKim: Not certain yet. I added the status codes to hopefully make it easier if ARINC wants to do something with them, and to provide a basic level of feedback for cases where we'd need to remote into a headend server and make an http request ourselves

lib/realtime_signs_web/controllers/monitoring_controller.ex
PaulJKim: Yeah, for now it's just the network topology that we're relying on.

Personally, I think access to the url is adequately restricted at the moment. If anything, since we're moving this value to S3, the value is arguably more openly and easily changeable via a direct upload to the bucket. I think adding at least a basic security layer for this endpoint is a good idea, but not sure how much we'd necessarily gain from that. Maybe we should add a separate task for setting up some authentication for the phoenix app as a whole though.

panentheos (approved):
Looks great, thanks for putting this together!

PR #600, opened 2023-03-27
Increase page durations for rail

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Stabilize paging frequency to address memory issues](https://app.asana.com/0/1201753694073608/1204168984928043/f)

This PR updates the few paging messages that we have to show each page for 4 seconds each (with the exception of JFK/UMass which we have set to page every 6 seconds) instead of 3 seconds. For background context: the "stops away" messages were set to show the first page for 6 seconds to convey that the primary piece of info is that the train is stopped. We decided internally to change this behavior so that each page shows for an equal duration.

The countdown clocks have been known to have their internal memory buffers overflow which causes crashes of unpredictable lengths.  Each page is sent by the SCU to the sign as an individual message at a frequency based on the page duration. By lengthening the page durations and making them more consistent, we hope to mitigate the potential for these memory-related crashes.

### DIFF
--- a/lib/content/message/alert/no_service_use_shuttle.ex
+++ b/lib/content/message/alert/no_service_use_shuttle.ex
@@ -18,12 +18,12 @@ defmodule Content.Message.Alert.NoServiceUseShuttle do
            PaEss.Utilities.destination_to_sign_string(destination),
            "no service",
            @default_page_width
-         ), 3},
+         ), 4},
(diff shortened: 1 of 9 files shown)


panentheos (approved):
There's one more paging message in `messages/predictions.ex` that might need addressing. After that, looks good.

PR #601, opened 2023-03-27
improve handling of bus prediction data

panentheos:
#### Summary of changes

* Uses API-provided headers to filter out responses that contain older data than we've gotten in previous responses. This fixes some instability in the prediction numbers over time.
* Includes trip id when identifying previous predictions. This accounts for the fact that we sometimes get predictions for future trips before the bus has passed on the current trip.
* Treats single-page messages as having a 6-second duration, for the purpose of delaying further updates. This will slow down sign updates in case we do encounter rapid changes in the predictions.


PaulJKim (approved):
One nit but otherwise LGTM!

PR #603, opened 2023-03-28
Add auth plug

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate adding security layer for monitoring endpoints](https://app.asana.com/0/1201753694073608/1204241391114683/f)

This PR adds a simple plug for any endpoints that realtime signs is hosting (for now this is just the monitoring endpoint) that will look for an api key in the headers and check against a corresponding value from its application config. This is to add a baseline level of security for these URLs.

We'll need to notify ARINC of the change before we deploy and share a key with them.

Once they update their scripts, we can populate our Secrets Manager environment(s) and deploy.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,8 +33,7 @@ config :realtime_signs,
   aws_client: ExAws,
   s3_client: ExAws.S3,
   number_of_http_updaters: 4,
-  restart_fn: &Engine.Health.restart_noop/0,
-  monitoring_api_key: "local_monitoring_api_key"
+  restart_fn: &Engine.Health.restart_noop/0
 
 config :realtime_signs, RealtimeSignsWeb.Endpoint, secret_key_base: "local_secret_key_base_at_least_64_bytes_________________________________"
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -34,7 +34,7 @@ config :realtime_signs,
   s3_client: ExAws.S3,
   number_of_http_updaters: 4,
   restart_fn: &Engine.Health.restart_noop/0,
-  monitoring_api_key: "local_monitoring_api_key",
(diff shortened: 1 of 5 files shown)


config/prod.exs
```diff
@@ -16,7 +16,8 @@ config :logger, :console, level: :info
 
 config :realtime_signs,
   external_config_getter: ExternalConfig.S3,
-  restart_fn: &System.restart/0
+  restart_fn: &System.restart/0,
+  monitoring_api_key: System.get_env("MONITORING_API_KEY")
```
panentheos, line 20: This should be set in `runtime.exs` to make sure it picks up the runtime environment instead of the build environment.

config/config.exs
```diff
@@ -33,7 +33,8 @@ config :realtime_signs,
   aws_client: ExAws,
   s3_client: ExAws.S3,
   number_of_http_updaters: 4,
-  restart_fn: &Engine.Health.restart_noop/0
+  restart_fn: &Engine.Health.restart_noop/0,
+  monitoring_api_key: "local_monitoring_api_key"
```
panentheos, line 37: Since this is security-related, I don't think we should provide a default value. It we set the config unconditionally from the environment, then it will be `nil` if the environment variable is missing, which should not be possible to get past (since header values are always strings). So the API would be inaccessible, which seems like a good default behavior.

panentheos (commented):
I tested this locally and it works as advertised, but see a few comments.

config/config.exs
PaulJKim: Hmm yeah that's a good point! Will remove this default

panentheos (approved):
Looks good!

PR #606, opened 2023-03-31
refactor headway code

panentheos:
#### Summary of changes

This aims to simplify the code that calculates headway messages. It does this by changing the structure of the signs JSON, and adding an explicit object under `source_config`, which holds the headway information alongside the prediction sources. Similar to before, platform signs have one config object, while mezzanine signs have a list of two. This allows for natural encodings of all existing cases, including the multiple headway groups at Ashmont. This results in simpler downstream logic for reading and interpreting the data.

Notes:
* The changes to `signs.json` were done entirely with the `transform_signs.exs` script, which is included here for review (but we can delete it afterward)
* This removes `source_for_headway`, since the headway configuration is now explicit and centralized. Currently that means the special multi-track stops (e.g. `Forest Hills-01`) will be included in the time-range calculation when deciding whether to show headways. I think this is ok, but if not, we can add logic to filter them out.
* There is one expected behavior change, the `park_st_westbound_wall_track` sign will now start showing the headway destination in its headway message. The fact that it wasn't already was likely a misconfiguration.
* The `wellington_southbound` sign will continue NOT showing a headway destination, but that fact is now encoded explicitly as a `null` in the new config object, rather than implicitly in the logic.
* Most of the headway messages now just include a `nil` as their associated source. This should not cause any behavior changes, because we only examine the source for prediction messages, but some of the tests needed to be adjusted.

### DIFF
--- a/lib/pa_ess/utilities.ex
+++ b/lib/pa_ess/utilities.ex
@@ -201,7 +201,8 @@ defmodule PaEss.Utilities do
   @doc """
   Used for parsing headway_direction_name from the source config to a PaEss.destination
   """
-  @spec headsign_to_destination(String.t()) :: {:ok, PaEss.destination()} | {:error, :unknown}
+  @spec headsign_to_destination(String.t()) ::
+          {:ok, PaEss.destination() | nil} | {:error, :unknown}
   def headsign_to_destination("Alewife"), do: {:ok, :alewife}
   def headsign_to_destination("Ashmont"), do: {:ok, :ashmont}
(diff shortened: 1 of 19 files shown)


PaulJKim (approved):
Great stuff here! Looked through everything and tested locally and seems good.

Re: Wellington Southbound, I think we can go ahead and remove the northbound source and define the southbound headway direction.

panentheos:
Thanks for the review! I'd like to make the Wellington Southbound change in a separate PR so it's easier for people to find in the future.

PR #608, opened 2023-04-03
enable bus sign engines

panentheos:
#### Summary of changes

This removes the `TEST_BUS_MODE` flag and enables the bus predictions and Chelsea bridge engines. These will fetch the relevant data in the background even though there are no enabled signs using them yet. After this, bus signs can be enabled one-by-one by adding entries to `signs.json`.

Also changes some code paths to behave correctly with empty data, and caches the list of bus stop ids on startup.

### DIFF
--- a/lib/realtime_signs.ex
+++ b/lib/realtime_signs.ex
@@ -21,11 +21,12 @@ defmodule RealtimeSigns do
         Engine.Departures,
         Engine.Static,
         Engine.Alerts,
+        Engine.BusPredictions,
+        Engine.ChelseaBridge,
         MessageQueue,
         RealtimeSigns.Scheduler,
         RealtimeSignsWeb.Endpoint
       ] ++
-        bus_children() ++
         http_updater_children() ++
         [
           Signs.Supervisor
@@ -64,16 +65,6 @@ defmodule RealtimeSigns do
 
     for i <- 1..num_children do
       {PaEss.HttpUpdater, i}
-    end
-  end
-
-  # These modules are specific to the in-progress bus work, and are disabled by default.
-  # They should be enabled and inlined once the work is complete.
-  def bus_children do
-    if Application.get_env(:realtime_signs, :test_bus_mode) do
-      [Engine.BusPredictions, Engine.ChelseaBridge]
-    else
-      []
     end
   end
 end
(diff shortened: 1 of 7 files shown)


test/signs/utilities/signs_config_test.exs
```diff
@@ -12,10 +12,4 @@ defmodule Signs.Utilities.SignsConfigTest do
       assert Enum.member?(Signs.Utilities.SignsConfig.all_train_stop_ids(), "70056")
     end
   end
-
-  describe "all_bus_stop_ids" do
```
PaulJKim, line 16: Would it be useful to just temporarily comment this test out and leave a to-do to add the test back so we don't forget about it?

lib/realtime_signs.ex
```diff
@@ -21,11 +21,12 @@ defmodule RealtimeSigns do
         Engine.Departures,
         Engine.Static,
         Engine.Alerts,
+        Engine.BusPredictions,
```
PaulJKim, line 24: Since these processes will be deployed in dev (and prod?) first without actually being used, should we add any logging so that we have a way to verify them as running? Or do you think it's reasonable to wait until Thursday to see

test/signs/utilities/signs_config_test.exs
panentheos: I'm not convinced of the usefulness of the other tests in this file, so I don't think we need to replicate them for the bus signs. I put this test case in there early on as a safeguard to ensure we didn't accidentally turn things on too early, but now that we're ready to deploy I think it's better to just remove it.

lib/realtime_signs.ex
PaulJKim: Hmm yeah that's fair. I don't think there's any reason to doubt that the processes will start up improperly so no need to add any logging there. And agreed, we probably don't want to add noise to the logs. I think we're good to just watch for a lack of errors and see how things go Thursday then!

lib/realtime_signs.ex
```diff
@@ -21,11 +21,12 @@ defmodule RealtimeSigns do
...
+        Engine.ChelseaBridge,
```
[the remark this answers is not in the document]
panentheos: That's a good question. I'm inclined to say that since the bridge is now an integral part of the app, those environment variables should be required. In any case I'll update the README so people know how to populate them.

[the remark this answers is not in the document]
PaulJKim: Sounds reasonable to me!

PaulJKim (approved):
LGTM! With the understanding that README will be updated with instruction on how to populate Chelsea Bridge API env vars for local development

PR #612, opened 2023-04-06
Increase message queue size to 400

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate increasing message queue size](https://app.asana.com/0/1201753694073608/1204331493928462/f)

Increase the message queue size to 400 to prevent a bit of overflow on startup and to leave some extra buffer room + room for the new bus signs that are to be added.


panentheos (approved):
Looks good!

PR #616, opened 2023-04-10
allow pointing at local sign config

panentheos:
#### Summary of changes

This leverages the recent signs-ui work to allow pointing at the config file of a local signs-ui. This enables a fully interactive local setup, where config changes in signs-ui are picked up and reflected in the sign output in real time, mirroring how the production system works.

Also removes support for the config file's "old format", which is no longer used.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -15,7 +15,6 @@ if config_env() != :test do
     s3_path: System.get_env("SIGNS_S3_PATH"),
     api_v3_url: System.get_env("API_V3_URL", "https://api-dev-green.mbtace.com"),
     api_v3_key: System.get_env("API_V3_KEY"),
-    sign_config_file: System.get_env("SIGN_CONFIG_FILE"),
     chelsea_bridge_url: System.get_env("CHELSEA_BRIDGE_URL"),
     chelsea_bridge_auth: System.get_env("CHELSEA_BRIDGE_AUTH"),
     filter_uncertain_predictions?: System.get_env("FILTER_UNCERTAIN_PREDICTIONS", "false") == "true",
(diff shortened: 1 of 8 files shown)


config/runtime.exs
```diff
@@ -15,6 +15,7 @@ if config_env() != :test do
     s3_path: System.get_env("SIGNS_S3_PATH"),
     api_v3_url: System.get_env("API_V3_URL", "https://api-dev-green.mbtace.com"),
     api_v3_key: System.get_env("API_V3_KEY"),
+    sign_config_file: System.get_env("SIGN_CONFIG_FILE"),
```
PaulJKim, line 18: Does `sign_config_file` need to be in the runtime config since it's relevant only for local development? In my mind, the runtime config is for stuff that we need to be configurable between deploying to dev or prod. Thoughts on moving it to `config.exs` or `dev.exs` and giving it a default of the value noted in the readme?

panentheos: That path is dependent on the directory structure that you have locally. I think it's probably common to have `realtime_signs` and `signs_ui` adjacent to each other, but I don't know if we should hardcode that. What about putting a comment on this line that calls it out as dev-only?

PaulJKim: Hmm yeah that's fair. I feel like setting a default isn't necessarily hard-coding it since a user could still use their own path in the same way, just with the extra step of having to re-compile but yeah I guess we don't necessarily want Realtime Signs to have an opinion about someone's local directory structure.

Another option could also be to make it more explicit and add a new section in `runtime.exs` for dev runtime config values by doing
```
if config_env() == :dev do
  sign_config_file: System.get_env("SIGN_CONFIG_FILE")
end
```

panentheos: Oh, that seems good, I'll implement that.

PaulJKim (approved):

PR #617, opened 2023-04-11
enable nubian A and C for bus signs cutover

panentheos:
#### Summary of changes

This enables Nubian platforms A and C for the bus signs cutover. It will be deployed alongside https://github.com/mbta/transitway_engine/pull/3

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -8376,6 +8376,80 @@
     "max_minutes": 30
   },
   {
+    "id": "bus.Nubian_Platform_A",
+    "pa_ess_loc": "SDUD",
+    "read_loop_interval": 420,
+    "read_loop_offset": 180,
+    "text_zone": "w",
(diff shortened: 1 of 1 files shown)


priv/signs.json
```diff
@@ -8375,6 +8375,80 @@
...
     "max_minutes": 30
   },
+  {
+    "id": "bus.Nubian_Platform_A",
+    "pa_ess_loc": "SDUD",
+    "read_loop_interval": 420,
+    "read_loop_offset": 180,
+    "text_zone": "w",
+    "audio_zones": [
+      "c"
```
PaulJKim, line 8385: Just double checking that these two platforms share the `c` audio zone?

panentheos: It looks like they do, yes. The `read_loop_offset` is different, so the readouts will be staggered.

PR #625, opened 2023-04-18
convert bus predictions engine to use ets

panentheos:
#### Summary of changes

This implements a possible solution to memory overflow issues that have been happening over the last few days. Converts the bus predictions engine to use ETS instead of process state for storage. This mirrors how the other engines typically handle data. The hypothesis is that the built-in copying that ETS does will allow the original large JSON strings to be garbage collected sooner, reducing memory pressure.


PaulJKim (approved):
LGTM!

PR #627, opened 2023-04-19
Increase memory

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate the erl crash](https://app.asana.com/0/1201753694073608/1204416582520667/f)

Doubles memory from 1024M to 2048M.

To be deployed after first trying in dev with #626


panentheos (approved):
Looks good, thanks!

PR #628, opened 2023-04-21
Always update both lines for rail

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Combine both lines into one message when both are paging](https://app.asana.com/0/1201753694073608/1204367042923156/f)

This change eliminates the use of single line updates and always the full sign any time the content on either of the lines changes. The aim of this is to simplify things and hopefully help the countdown clocks perform a bit more stably. 

This change also makes it so that any time both lines of the sign are paging separately (i.e. the top line paging "stops away" and the bottom line paging headways) the request to page both lines will be sent as a single request to the headend server. We have been observing signs consistently crashing when both lines are paging at the same time so we are hoping that sending the update as a single request will allow the sign to page properly without crashing.

Note: If this PR is merged, it means that we can remove all of the logic related to updating a single line at a time. I decided not to overload this PR with all of those changes and we can follow up with another PR to remove all of that logic when needed.


lib/signs/utilities/updater.ex
```diff
@@ -19,43 +19,27 @@ defmodule Signs.Utilities.Updater do
...
-
-        %{sign | current_content_bottom: bottom_msg, tick_bottom: sign.expiration_seconds}
-
-      # update both
-      {false, false} ->
-        log_line_update(sign, top_msg, "top")
-        log_line_update(sign, bottom_msg, "bottom")
+      _ ->
+        top_msg =
+          if Messages.same_content?(sign.current_content_top, top_msg) do
```
panentheos, line 29: nit: Since these `same_content?` calls are non-trivial and central to the logic here, it might be nice do do them once at the beginning and store the results in variables, so it's clear at a glance that the same values are being used further down.

panentheos (approved):
See minor comment, then looks good. Tested locally. I like how this simplifies things!

lib/signs/utilities/updater.ex
PaulJKim: Ah yeah, that's a good call!

PR #629, opened 2023-04-24
remove old logging mode

panentheos:
#### Summary of changes

This removes the `DRY_RUN` environment variable and associated logging code. If desired, an improved version of this can be achieved by simply unsetting the environment variables for the headend server and signs-ui.


PaulJKim (approved):
LGTM!

PR #630, opened 2023-04-25
Remove single line logic

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Remove logic related to updating single lines for rail](https://app.asana.com/0/1201753694073608/1204460989217745/f)

Since we made messages always update both lines at the same time with #628 , we can remove all of the single line update logic.


panentheos (approved):
Looks good!

PR #631, opened 2023-04-25
show no service message for bus alerts

panentheos:
#### Summary of changes
**Asana Ticket:** [Play "No bus service" at busways/transitway stops when all service is eliminated by alerts](https://app.asana.com/0/1201753694073608/1204324618353134/f)

This adds basic alert handling for bus signs. When no predictions are present, and the alert level indicates that the stop is closed, display and read "No bus service" instead of just showing a blank sign.


PaulJKim (approved):
Tested locally and looks good!

PR #632, opened 2023-04-26
use direnv for local environment management

panentheos:
#### Summary of changes

This updates the documentation to specify using direnv to manage environment variables locally. This gives us a convenient way to include detailed inline documentation, and to suggest some values for wiring up the local environment without hard-coding them. Notes:
* Removed documentation for `SIGN_HEAD_END_HOST`, since this is handled by the config module now.
* Removed documentation for `NUMBER_OF_HTTP_UPDATERS`, on the premise that it's almost never necessary to modify this in practice.
* Removed the default value for `API_V3_URL` in `runtime.exs`, for consistency. Confirmed that both deployed environments will supply this via environment variables.


PaulJKim (approved):
Looks great! Thank you for doing this!

PR #633, opened 2023-04-28
centralize and deduplicate prediction fetching

panentheos:
#### Summary of changes

This extracts the rail prediction fetching to the top of the sign loop, and passes the predictions down to code paths that need them. This has two benefits:
* The data flow is clearer, due to centralizing access to external data sources.
* The predictions are only fetched once (rather than twice), so it's a bit more efficient.

### DIFF
--- a/lib/signs/utilities/predictions.ex
+++ b/lib/signs/utilities/predictions.ex
@@ -9,35 +9,34 @@ defmodule Signs.Utilities.Predictions do
   require Content.Utilities
   alias Signs.Utilities.SourceConfig
 
-  @spec get_messages(Signs.Realtime.t()) :: Signs.Realtime.sign_messages()
+  @spec get_messages(Signs.Realtime.predictions(), Signs.Realtime.t()) ::
+          Signs.Realtime.sign_messages()
   def get_messages(
-        %{
-          source_config: {%{sources: top_line_sources}, %{sources: bottom_line_sources}},
-          text_id: {station_code, zone}
-        } = sign
+        {top_predictions, bottom_predictions},
+        %{source_config: {top_config, bottom_config}} = sign
       ) do
-    {top, _} = get_predictions(sign.prediction_engine, top_line_sources, station_code, zone)
-    {bottom, _} = get_predictions(sign.prediction_engine, bottom_line_sources, station_code, zone)
+    {top, _} = prediction_messages(top_predictions, top_config, sign)
+    {bottom, _} = prediction_messages(bottom_predictions, bottom_config, sign)
 
     {top, bottom}
   end
 
-  def get_messages(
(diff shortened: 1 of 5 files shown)


test/signs/utilities/predictions_test.exs
panentheos, line 1: TOL: Fixing the test code took more than 50% of the time and effort in making this change, which seems unfortunate for a refactor that shouldn't affect externally facing behavior. If these test cases were written against the top-level sign interface instead of testing internal functions, I don't think that would have been the case.

test/signs/utilities/messages_test.exs
```diff
@@ -13,58 +13,54 @@ defmodule Signs.Utilities.MessagesTest do
...
+      route_id: "Red",
+      stopped?: false,
+      stops_away: 1,
+      destination_stop_id: "70093",
+      seconds_until_arrival: 240,
+      seconds_until_departure: 300
+    }
+  ]
+
+  @nd_predictions [
```
PaulJKim, line 43: Is `@nd_predictions` a typo or does `nd` stand for something?

lib/signs/utilities/predictions.ex
```diff
@@ -99,36 +98,24 @@ defmodule Signs.Utilities.Predictions do
     end
   end
 
-  @spec get_passthrough_train_audio(Signs.Realtime.t()) :: [Content.Audio.t()]
-  def get_passthrough_train_audio(
-        %Signs.Realtime{source_config: %{sources: single_source}} = sign
-      ) do
-    single_source |> get_source_list_passthrough_audio(sign.prediction_engine) |> List.wrap()
+  @spec get_passthrough_train_audio(Signs.Realtime.predictions()) :: [Content.Audio.t()]
```
PaulJKim, line 101: Wow nice. This whole section of code is now much more expressive!

test/signs/utilities/predictions_test.exs
PaulJKim: Hmm yeah I agree it's a pain and creates some unnecessary rigidity against doing refactors like this one. I think having some unit tests for each module, including more internal ones is helpful though. But maybe there doesn't need to be a 1000+ line test file for an internal function like in this case and the tests should be more focused at the top-level as you said.

PaulJKim (approved):
Had one inconsequential question, but otherwise looks good!

test/signs/utilities/messages_test.exs
panentheos: It stands for "no departures". I had spelled it out originally, but that caused the auto-formatter to splay out a whole bunch of lines below 😠 .

test/signs/utilities/predictions_test.exs
panentheos: Yep, there's probably a balance to be struck there.

PR #636, opened 2023-05-08
Increase page duration to 5 seconds

PaulJKim:
#### Summary of changes
**Asana Ticket:** [ARINC investigates crashes seemingly caused by MZ dual-paging](https://app.asana.com/0/1201753694073608/1204553377557896/f)

This PR increases page durations to 5 seconds from 4 for rail-related paging messages. This is an effort to prevent signs from crashing when both lines of the sign are paging at the same time.  In the past when the JFK/UMass sign exhibited similar behavior, increasing the page duration stabilized the sign.

If we observe that signs still crash when the paging is at a 5 second interval, we can try 6. After 6 though, I feel that the pages may be getting too long and we'd probably want to consider other solutions.


panentheos (approved):
Looks good!

PR #637, opened 2023-05-09
Increase page durations to 6 seconds

PaulJKim:
#### Summary of changes
**Asana Ticket:** [ARINC investigates crashes seemingly caused by MZ dual-paging](https://app.asana.com/0/1201753694073608/1204553377557896/f)

Increase page durations again to 6 seconds to see if this stabilizes double line paging for rail signs.

Now that pages are getting pretty long, we should stop increasing the durations and rely on ARINC addressing the issue at the SCU level if this doesn't work.


PaulJKim:
woops forgot to merge main first 🤦 addressing the conflicts now

panentheos (approved):
It looks like there's a `format_bottom` function in `headway_display.ex` that's generating a 5-second paging interval, which we might want to also change. Aside from that, looks good.

PR #638, opened 2023-05-11
Add sign id to logs and make content more human-readable

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Add human readable logging to RTS messages for Transit Data](https://app.asana.com/0/1201753694073608/1204564881899172/f)

This PR introduces a change to pass the `sign_id` along so that we can log it when we send requests to ARINC. This is so that we can more easily query for requests that went to specific sign zones.

It also adds new fields for the top and bottom content to make it easier to parse out what was sent to a sign zone.

### DIFF
--- a/lib/message_queue.ex
+++ b/lib/message_queue.ex
@@ -35,16 +35,16 @@ defmodule MessageQueue do
   end
 
   @impl PaEss.Updater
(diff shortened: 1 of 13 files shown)


panentheos (approved):
Looks good! Tested locally.

PR #639, opened 2023-05-12
send audio messages to signs-ui

panentheos:
#### Summary of changes
**Asana Ticket:** [Signs-UI shows a flag/toggle when audio has been sent to a zone](https://app.asana.com/0/1201753694073608/1204564881899174/f)

This sends audio messages to signs_ui, now that it can accept them. It consolidates the code paths for posting and logging messages, and fixes a case where we should have been logging `sign_id` but weren't.


PaulJKim (approved):
LGTM! And thank you for cleaning up the logging logic while you're at it!

PR #642, opened 2023-06-05
add historical prediction analysis script

panentheos:
#### Summary of changes

This adds a script for investigating historical prediction values. It pulls raw data for the specified time period, performs the same kind of first-pass filtering as the app, and outputs a CSV showing a minute-by-minute breakdown of the closest prediction value for the given signs. This is useful to visualize the raw numbers we're receiving, to inform how we should transform them further.


PaulJKim (approved):
Was able to run locally and looks good to me!

PR #644, opened 2023-06-14
Clean up stops.json and add new stops for B/E branches

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate Babcock alert treatment](https://app.asana.com/0/1201753694073608/1204800787143423/f)

Stemming from the above linked investigation, this PR adds a few new stops to the `stops.json` file for the B branch so that alerts that start between Kenmore and Amory get applied correctly to the countdown clocks as well as Babcock. Similarly add Northeastern so that Symphony gets proper treatment.

It also cleans up the file a bit and consolidates some of the lists. The lists for Green E and GLX don't need to be in the same list in order to work as intended.

### DIFF
--- a/priv/stops.json
+++ b/priv/stops.json
@@ -25,31 +25,29 @@
     {"station": "G Hynes westbound", "stop_ids": ["70153"]},
     {"station": "G Kenmore westbound", "stop_ids": ["70151"]}
   ],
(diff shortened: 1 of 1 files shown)


cmaddox5 (approved):
🚀

PR #650, opened 2023-07-03
unify sign update code paths

panentheos:
#### Summary of changes

This rearranges the sign update logic so that both content- and expiration-related updates happen via the same unified code path. It also measures expiration using timestamps directly, rather than indirectly counting "ticks" between updates.


PaulJKim (approved):
This is great!

PR #656, opened 2023-07-18
Add other GL routes to Medford branch source configs

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Allow B, C, D predictions to show on Medford branch signs](https://app.asana.com/0/1201753694073608/1205057928541598/f)

Currently, any B, C, or D predictions get filtered out before they make it onto Medford branch signs. While these trips don't happen frequently, there are usually a few per day so we should let the predictions show on the countdown clocks.

[Slack thread](https://mbta.slack.com/archives/C03K6NLKKD1/p1689263243534829) for context

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -7807,6 +7807,9 @@
           "stop_id": "70513",
           "direction_id": 1,
           "routes": [
+            "Green-B",
+            "Green-C",
+            "Green-D",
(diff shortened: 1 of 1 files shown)


panentheos (approved):
Looks good. Did not do additional testing since it's just a config change.

PR #657, opened 2023-07-18
Log route-based subway predictions filtering

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate removing route-based filtering for predictions](https://app.asana.com/0/1201753694073608/1205057928541600/f)

This filtering appears to not be useful since there are always distinct platform stop ids for the distinct routes at a given station. We want to see how often, if at all, predictions are getting filtered out by route with this temporary logging.


panentheos (approved):
Looks good.

PR #658, opened 2023-07-19
Add takes for general headway directions

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Implement early AM prediction changes](https://app.asana.com/0/1201753694073608/1204795336662073/f)

While implementing early AM logic, the need for having pre-recorded takes for Southbound, Northbound, etc. came up so that we could use pre-recorded takes in order to read out times (e.g. 5:03 as "five-oh-three"). I'm not certain if ARINC's TTS would handle reading times correctly and we had the takes for time readouts already so it was simple enough to just add a couple more takes for the general headway directions instead.

This had an unintended side-affect of causing all normal prediction-style readouts from using the new takes as well rather than defaulting to an ad-hoc TTS readout. This would result in readouts such as "The next train to Southbound arrives in 5 minutes".

We could rework the audio logic to use the new takes and switch up the wording, but that seemed out of scope of the original task so instead, this PR just adds the new takes but maintains the status quo of doing ad-hoc readouts for general headway directions for existing message types.

### DIFF
--- a/lib/content/audio/following_train.ex
+++ b/lib/content/audio/following_train.ex
@@ -60,36 +60,51 @@ defmodule Content.Audio.FollowingTrain do
     def to_params(audio) do
       case Utilities.destination_var(audio.destination) do
         {:ok, dest_var} ->
-          green_line_branch = Content.Utilities.route_branch_letter(audio.route_id)
+          if audio.destination in [
+               :southbound,
+               :northbound,
+               :westbound,
+               :eastbound,
+               :inbound,
+               :outbound
+             ] do
+            do_ad_hoc_message(audio)
+          else
+            green_line_branch = Content.Utilities.route_branch_letter(audio.route_id)
 
-          cond do
-            !is_nil(green_line_branch) ->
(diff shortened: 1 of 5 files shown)


lib/content/audio/following_train.ex
```diff
@@ -60,36 +60,51 @@ defmodule Content.Audio.FollowingTrain do
...
-            true ->
-              {:canned, {"160", [dest_var, verb_var(audio), minutes_var(audio)], :audio}}
+          if audio.destination in [
+               :southbound,
+               :northbound,
+               :westbound,
+               :eastbound,
+               :inbound,
+               :outbound
+             ] do
```
panentheos, line 70: Since we're doing this check multiple times, what about creating a helper, e.g. `directional_destination?`

panentheos (approved):
See comment, then looks good.

lib/content/audio/following_train.ex
PaulJKim: Yeah good point!

PR #659, opened 2023-07-20
Add Kenmore B stops as destinations

PaulJKim:
#### Summary of changes
Ad-hoc

Kenmore B stops were not included in the mapping from destination stop_ids to destinations so B branch predictions were defaulting to using `Boston Col` when the destination stop id was being calculated as `71151`. Went ahead and added the EB Kenmore platform id for the future.


jfabi (approved):
Looks good to me! 🐝

PR #660, opened 2023-07-20
Early AM predictions suppression content

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Implement early AM prediction changes](https://app.asana.com/0/1201753694073608/1204795336662073/f)

Add new content structs that will be used for early AM predictions suppreesion and related helpers/unit tests

### DIFF
--- a/lib/content/message/early_am/destination_scheduled_time.ex
+++ b/lib/content/message/early_am/destination_scheduled_time.ex
@@ -12,7 +12,7 @@ defmodule Content.Message.EarlyAm.DestinationScheduledTime do
           destination: destination,
           scheduled_time: scheduled_time
         }) do
-      "#{String.capitalize(PaEss.Utilities.destination_to_sign_string(destination))} due #{scheduled_time.hour}:#{Content.Utilities.format_minutes(scheduled_time.minute)}"
+      "#{String.capitalize(PaEss.Utilities.destination_to_sign_string(destination))} due #{Content.Utilities.render_datetime_as_time(scheduled_time)}"
     end
   end
 end
--- a/lib/content/message/early_am/destination_scheduled_time.ex
+++ b/lib/content/message/early_am/destination_scheduled_time.ex
@@ -1,0 +1,18 @@
+defmodule Content.Message.EarlyAm.DestinationScheduledTime do
+  @enforce_keys [:destination, :scheduled_time]
+  defstruct @enforce_keys
+
+  @type t :: %__MODULE__{
+          destination: PaEss.destination(),
+          scheduled_time: DateTime.t()
+        }
+
+  defimpl Content.Message do
+    def to_string(%Content.Message.EarlyAm.DestinationScheduledTime{
+          destination: destination,
+          scheduled_time: scheduled_time
+        }) do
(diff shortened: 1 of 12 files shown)


lib/signs/utilities/audio.ex
```diff
@@ -281,6 +281,40 @@ defmodule Signs.Utilities.Audio do
...
+       ) do
+    Audio.FirstTrainScheduled.from_messages(top, bottom)
+  end
+
+  defp get_audio(
+         %Message.Predictions{station_code: "RJFK"} = top_content,
+         %Message.PlatformPredictionBottom{},
+         multi_source?
+       ) do
+    Audio.Predictions.from_sign_content(%{top_content | zone: "m"}, :bottom, multi_source?)
```
panentheos, line 315: This line is confusing to me. Why are we setting the zone? And do we mean to pass `top_content` with `:bottom`?

PaulJKim: Ah yeah this probably doesn't make sense without taking into context the subsequent PR, which perhaps this change should have gone with.

In order to make the RJFK platform prediction render "normally", aka without the platform stuff, so that we can tag the single-line platform message to the bottom, I [changed the prediction struct to omit the zone](https://github.com/mbta/realtime_signs/pull/661/files#diff-bdda4cc40319158f7ed70a48350c960125e1cc22848b0c5f9dccde12c05205beR90) so that it wouldn't trigger that logic. With the audio though, we _do_ want the platform logic to apply so I add back the zone here to make that happen.

In this specific case, it actually doesn't matter what we pass in for the second parameter. The only condition in the called method that depends on line checks for `line == :top or multi_source?` and at this point we know it's a multi-source sign so it would evaluate to `true` regardless. I used `:bottom` because _technically_ the alewife sources are the bottom ones for JFK/UMass, but it was just to pass something for that parameter.

lib/content/message/early_am/destination_scheduled_time.ex
```diff
@@ -0,0 +1,18 @@
...
+          destination: PaEss.destination(),
+          scheduled_time: DateTime.t()
+        }
+
+  defimpl Content.Message do
+    def to_string(%Content.Message.EarlyAm.DestinationScheduledTime{
+          destination: destination,
+          scheduled_time: scheduled_time
+        }) do
+      "#{String.capitalize(PaEss.Utilities.destination_to_sign_string(destination))} due #{scheduled_time.hour}:#{Content.Utilities.format_minutes(scheduled_time.minute)}"
```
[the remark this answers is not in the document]
PaulJKim: Hmm right, yeah I think they did that with alerts data especially. I definitely can see the benefit of that separation after working on the early AM logic and figuring out how to handle rendering the content depending on the combination of messages and whether it was a mezzanine or not. Certainly think that conversation would be worthwhile, especially with the future of PA/ESS in mind

lib/signs/utilities/audio.ex
panentheos: Got it. That makes sense, but I think we should include a comment explaining the zone override, and the linkage between this and the other code. (I think it'll be non-obvious even when both pieces are in)

panentheos (approved):
See followup comment, then looks good!

PR #661, opened 2023-07-20
JFK umass mezzanine refactor

PaulJKim:
#### Summary of changes

The JFK/UMass Mezzanine case where the top line (SB) is on headways while the bottom line is on a prediction was not handled in the previous version and would try to page the prediction on the top line while showing paging headways on the bottom line. This would surpass the character limit for the top line.

This PR takes advantage of some of the new code added for the Early AM Predictions suppression to handle this case.

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -14,34 +14,98 @@ defmodule Signs.Utilities.Messages do
           Engine.Alerts.Fetcher.stop_status()
         ) :: Signs.Realtime.sign_messages()
   def get_messages(predictions, sign, sign_config, current_time, alert_status) do
-    cond do
-      match?({:static_text, {_, _}}, sign_config) ->
-        {:static_text, {line1, line2}} = sign_config
+    messages =
+      cond do
+        match?({:static_text, {_, _}}, sign_config) ->
+          {:static_text, {line1, line2}} = sign_config
 
-        {Content.Message.Custom.new(line1, :top), Content.Message.Custom.new(line2, :bottom)}
+          {Content.Message.Custom.new(line1, :top), Content.Message.Custom.new(line2, :bottom)}
 
-      sign_config == :off ->
-        {Content.Message.Empty.new(), Content.Message.Empty.new()}
+        sign_config == :off ->
+          {Content.Message.Empty.new(), Content.Message.Empty.new()}
 
-      sign_config == :headway ->
-        get_headway_or_alert_messages(sign, current_time, alert_status)
+        sign_config == :headway ->
+          get_headway_or_alert_messages(sign, current_time, alert_status)
 
-      true ->
-        case Signs.Utilities.Predictions.get_messages(predictions, sign) do
-          {%Content.Message.Empty{}, %Content.Message.Empty{}} ->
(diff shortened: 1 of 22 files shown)


lib/signs/utilities/messages.ex
```diff
@@ -14,34 +14,99 @@ defmodule Signs.Utilities.Messages do
...
+           sign,
+           current_time,
+           alert_status,
+           :top
+         ) do
+      %Content.Message.Headways.Paging{destination: destination, range: range} ->
+        {%Content.Message.GenericPaging{
+           messages: [
+             # Make zone nil in order to prevent the usual paging platform message
+             %{prediction | zone: nil},
```
panentheos, line 91: With this prediction message nested inside the generic paging message, the `countup?` logic won't trigger as usual. We need to expand that logic if we want it to keep working for this case.

panentheos (approved):
See comments, otherwise looks good.

lib/signs/utilities/messages.ex
PaulJKim: Ah great catch, thank you!

lib/signs/utilities/messages.ex
```diff
@@ -14,34 +14,99 @@ defmodule Signs.Utilities.Messages do
...
+              end
+
+            messages ->
+              messages
+          end
+      end
+
+    flip? = flip?(messages)
+
+    if flip?,
```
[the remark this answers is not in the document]
PaulJKim: I may leave this one for now in as I re-use the `flip?` variable in a couple more places in the subsequent PR

lib/signs/utilities/messages.ex
```diff
@@ -14,34 +14,99 @@ defmodule Signs.Utilities.Messages do
...
+                 ) do
+                jfk_umass_headway_paging(bottom_message, sign, current_time, alert_status)
+              else
+                {get_paging_headway_or_alert_messages(sign, current_time, alert_status, :top),
+                 bottom_message}
+              end
+
+            messages ->
+              messages
+          end
```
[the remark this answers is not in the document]
PaulJKim: Agreed although, I don't think I have as clear of an idea how that different approach would look in practice. I'd like to hear in more detail about how you're envisioning things! Let's chat soon about it

lib/signs/utilities/messages.ex
PaulJKim: @panentheos just pushed some changes to expand the countup logic. Mind taking another quick pass? 🙏

PaulJKim: Ohh hm yes, I think that's correct. We can't match up the predictions when we're transitioning between the two formats and would potentially miss a count-up.

Alternatively, I think we could get around having to do a more substantial refactor by mapping the `PlatformPredictionBottom` struct to a regular `Prediction` struct before calling `countup?` again. I think it should be a safe assumption that the destinations are the same (`:alewife`) since this is such a specific case, or we could also just pass along the `destination` when we create the `PlatformPredictionBottom` struct. We already pass along the minutes to the bottom so we would have all the pieces handy for the countup logic to do its thing.

panentheos: Yeah, I guess that works, as long as we call out with a comment that it's tailored for the special case.

PR #665, opened 2023-08-01
test message logic at the sign API level

panentheos:
#### Summary of changes

This reimplements the suite of "message" tests as top-level sign tests. Benefits:
* The test cases are smaller and simpler, with clearly visible inputs and outputs.
* They exercise the same underlying logic, but in a more "real" way, since the full sign code path is called.
* This will ultimately allow us to refactor the underlying code paths if desired, while verifying that the overall system behavior hasn't changed.

The non-test changes are needed to support the use of Mox for mocking external calls.


PaulJKim (approved):
LGTM! Would it be possible to wait to merge this until after the Early AM stuff gets merged it? There are some new tests added to `messages_test.exs` in that PR that will need to be moved/updated too.

Feel free to go ahead and merge if you'd like though, since I'll be out for the next week. I'll just update the early AM PR in that case

PR #670, opened 2023-08-14
Add logic to compare to scheduled first arrival

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Implement early AM prediction changes](https://app.asana.com/0/1185117109217413/1204795336662073/f)

Fetching headways in RTS has some embedded logic that will fetch the first and last scheduled arrivals and compare the current time against them to see if headways should be shown or not. This logic is performed at the **sign** level rather than the source level.

In the early AM hours, this sign-level logic is causing both lines/sources of mezzanine signs to go to headways when one of the lines should have stayed on a timestamp message.

This PR adds some additional logic that does a check of the current time + upper headway range against the scheduled first arrival for the specific sources. If the headways should be active, then the logic passes on the headway messages. If the headways should not be active, it will default back to the timestamp message.

### DIFF
--- a/lib/signs/utilities/early_am_suppression.ex
+++ b/lib/signs/utilities/early_am_suppression.ex
@@ -202,9 +202,19 @@ defmodule Signs.Utilities.EarlyAmSuppression do
                    scheduled_time: scheduled
                  }}
 
-              {headway_top, headway_bottom} ->
-                # Make sure routes is nil so that destination is used as headsign
-                {%{headway_top | destination: destination, routes: nil}, headway_bottom}
+              {headway_top, %Headways.Bottom{range: {_, high}} = headway_bottom} ->
+                # Check against schedule to account for mezzanine sources having different first scheduled arrivals
+                if DateTime.add(current_time, high, :minute) |> DateTime.compare(scheduled) == :gt do
+                  # Make sure routes is nil so that destination is used as headsign
+                  {%{headway_top | destination: destination, routes: nil}, headway_bottom}
+                else
+                  {%EarlyAm.DestinationTrain{
+                     destination: destination
+                   },
+                   %EarlyAm.ScheduledTime{
+                     scheduled_time: scheduled
+                   }}
+                end
             end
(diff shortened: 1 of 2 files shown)


lib/signs/utilities/early_am_suppression.ex
```diff
@@ -202,9 +202,19 @@ defmodule Signs.Utilities.EarlyAmSuppression do
                    scheduled_time: scheduled
                  }}
 
-              {headway_top, headway_bottom} ->
-                # Make sure routes is nil so that destination is used as headsign
-                {%{headway_top | destination: destination, routes: nil}, headway_bottom}
+              {headway_top, %Headways.Bottom{range: {_, high}} = headway_bottom} ->
+                # Check against schedule to account for mezzanine sources having different first scheduled arrivals
+                if DateTime.add(current_time, high, :minute) |> DateTime.compare(scheduled) == :gt do
```
panentheos, line 207: Just to check my understanding: This wouldn't be susceptible to the opposite problem, showing a timestamp when we should show headways, because the headways code takes the earliest time from all the stops for the sign. So it will always be available when we need it, but sometimes we need to suppress it again for 1/2 of the mezzanine sign. Right?

panentheos (approved):
Looks good, see question.

lib/signs/utilities/early_am_suppression.ex
PaulJKim: Correct, headways will always be returned based on the earlier first scheduled arrival between the two source configs. We'll only sometimes have a need to suppress headways for the half associated with the later first scheduled arrival when that other half is less than 40 min from the first scheduled arrival, but more than [upper headway range] min from the first first scheduled arrival **and** there are no valid predictions to be shown.

This is probably a rather rare case on heavy rail since trains are staged long before the first scheduled arrivals resulting in terminal predictions. It's going to be more common on the GL, in particular GLX stations, since we don't have terminal predictions there yet.

PR #672, opened 2023-08-15
Add new engine for locations feed

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🟠🧑‍🤝‍🧑🔊 Investigate Location Data Feed](https://app.asana.com/0/1185117109217413/1205194882142437/f)

Currently, the Predictions engine downloads the locations feed and immediately passes it on to the departure engine which is no longer used.

In light of locations data being needed in the near future both for OL Crowding data as well as headway tracking, this PR sets up a new engine that downloads and maintains a table of locations data keyed on vehicle_ids.

Note: This is data is coming from RTR so contains only subway data. No bus location is being read from this feed.


lib/locations/location.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Locations.Location do
+  defstruct vehicle_id: nil,
+            status: nil,
+            stop_id: nil,
+            timestamp: nil,
+            route_id: nil,
+            trip_id: nil,
+            consist: []
```
PaulJKim, line 8: I assume this will change/be added to. I just picked a general set of fields to start us off with

lib/engine/locations.ex
```diff
@@ -0,0 +1,150 @@
...
+
+  @impl true
+  def handle_info(:update, state) do
+    schedule_update(self())
+
+    last_modified_vehicle_locations =
+      download_and_process_vehicle_locations(
+        state[:last_modified_vehicle_positions],
+        :vehicle_positions_url,
+        state[:vehicle_locations_table]
```
panentheos, line 53: nit: could use dot notation on `state` here, and there's no good reason to pass `:vehicle_positions_url` as a parameter, we could just inline it in the function.

lib/engine/locations_api.ex
```diff
@@ -0,0 +1,3 @@
+defmodule Engine.LocationsAPI do
+  @callback for_vehicle(String.t()) :: [Predictions.Prediction.t()]
```
panentheos, line 2: This should return a single `Location.Location.t`

panentheos (approved):
See comments, then looks good.

PR #675, opened 2023-08-18
add departure logging for headway analysis

panentheos:
#### Summary of changes
**Asana Ticket:** [Track & flag headway accuracy in real time](https://app.asana.com/0/1185117109217413/1205116629331715/f)

This adds real-time logging of departures at specified stops, in order to measure headway accuracy. The actual calculations and monitoring will be done in Splunk.

Also removes the deprecated Departures engine, and the previous headway accuracy logging.


lib/engine/locations.ex
```diff
@@ -28,51 +30,52 @@ defmodule Engine.Locations do
...
   end
 
+  @impl true
+  def for_stop(stop_id) do
+    case :ets.lookup(@stop_locations_table, stop_id) do
+      [{^stop_id, locations}] -> locations
+      _ -> []
+    end
+  end
+
```
PaulJKim, line 40: Mainly out of curiosity, did you consider the approach of querying the ETS table by stop_id values within the `vehicle_locations_table`? I haven't played around with ETS all that much, but I think something like that should be doable?

lib/headway_analysis/server.ex
```diff
@@ -0,0 +1,75 @@
+defmodule HeadwayAnalysis.Server do
```
PaulJKim, line 1: Why not just start tracking vehicles present and doing the logging in the `Signs.Realtime` struct instead of setting up these new servers?

panentheos: I thought about that. Since this seems like an orthogonal task that's independent from what the main sign processes do, I thought it would be cleaner to put the code in its own processes.

PaulJKim: Gotcha. Makes sense to me

PaulJKim (approved):
LGTM!

PR #677, opened 2023-08-23
Determine emptier car location and crowding description

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🟠🧑‍🤝‍🧑🔊 Crowding Message Selection Logic](https://app.asana.com/0/1185117109217413/1205194882142439/f)

This PR adds logic that calculates the emptier car locations and crowding description. These pieces will later be used to 
create the full audio message that will get tagged onto either the Arrival or Approaching message.

The "crowding zone" logic is depicted in [this diagram](https://miro.com/app/board/uXjVMtnCjPc=/?moveToWidget=3458764561696176899&cot=14)


lib/content/message/predictions.ex
```diff
@@ -154,17 +196,101 @@ defmodule Content.Message.Predictions do
...
+  end
+
+  defp in_zone?(_, _, nil), do: false
+
+  defp in_zone?(car_nums, zone, {front_end, back_end}) do
+    special_case? =
+      (Enum.count(car_nums) == 2 and zone == :front and car_nums == [1, 4]) or
+        (Enum.count(car_nums) == 2 and zone == :back and car_nums == [3, 6]) or
+        (Enum.count(car_nums) == 4 and zone == :front and car_nums == [1, 2, 3, 6]) or
+        (Enum.count(car_nums) == 4 and zone == :back and car_nums == [1, 4, 5, 6])
```
panentheos, line 282: Since we're doing a full equality test on `car_nums`, it looks like the `count` checks are redundant

lib/content/message/predictions.ex
```diff
@@ -154,17 +196,101 @@ defmodule Content.Message.Predictions do
...
+
+    special_case? or
+      Enum.all?(car_nums, fn car_num -> car_num in Enum.to_list(front_end..back_end) end)
```
panentheos, line 285: nit: We can test for membership directly in a range, without converting it to a list. E.g. `car_num in front_end..back_end`

lib/content/message/predictions.ex
```diff
@@ -154,17 +196,101 @@ defmodule Content.Message.Predictions do
...
+  end
+
+  defp in_zone?(car_nums, _zone, zone_patterns) when is_list(zone_patterns) do
+    Enum.any?(zone_patterns, fn pattern ->
+      car_nums == pattern
+    end)
```
panentheos, line 291: nit:
```suggestion
    car_nums in zone_patterns
```

lib/content/message/predictions.ex
```diff
@@ -49,9 +50,55 @@ defmodule Content.Message.Predictions do
...
+    5 => [
+      front: {1, 5},
+      back: {2, 6},
+      middle: nil,
+      train_level: [[1, 3, 4, 5, 6], [1, 2, 4, 5, 6], [1, 2, 3, 5, 6], [1, 2, 3, 4, 6]]
+    ],
+    6 => [
+      train_level: {1, 6}
+    ]
+  }
```
panentheos, line 95: Thinking out loud: This could be a place where pattern matching really shines. If we create a list that represents the status of each car, e.g. `[:f, :e, :e, :f, :f, :f]`, we could pattern match against that, along with the number of empty cars, and create a series of cases like: `{2, [_, _, _, :f, :f, :f]} -> :front`. That gives a nice visual way to express the logic, and would also allow us to fold in the special cases from the code below.

PaulJKim: Hmm yeah that's an interesting idea. I sort of like how the current approach condenses the logic but I can also see that it's a bit more obscure than expressing everything as patterns. Will give it a try and see how it turns out

lib/content/message/predictions.ex
```diff
@@ -154,17 +151,100 @@ defmodule Content.Message.Predictions do
...
+    case crowding_level do
+      1 -> :not_crowded
+      2 -> :some_crowding
+      3 -> :crowded
+      4 -> :unknown_croding
+    end
+  end
+
+  defp get_emptier_location(car_crowding_levels) do
+    case car_crowding_levels do
```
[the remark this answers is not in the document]
panentheos: I really like this. It's very dense, but everything is laid out so it's very traceable. Also the simplification of the caller is pretty huge.

lib/content/message/predictions.ex
```diff
@@ -154,17 +151,100 @@ defmodule Content.Message.Predictions do
...
+      {1, [_, _, :f, :f, :f, :f]} -> :front
+      {1, [:f, :f, :f, :f, _, _]} -> :back
+      {1, [:f, :f, _, _, :f, :f]} -> :middle
+      {2, [_, _, _, :f, :f, :f]} -> :front
+      {2, [_, :f, :f, _, :f, :f]} -> :front
+      {2, [:f, :f, :f, _, _, _]} -> :back
+      {2, [:f, :f, _, :f, :f, _]} -> :back
+      {2, [:f, :f, _, _, :f, :f]} -> :middle
+      {2, [:f, _, :f, _, :f, :f]} -> :middle
+      {2, [:f, :f, _, :f, _, :f]} -> :middle
```
panentheos, line 227: Looks like `2` needs case(s) for `:front_and_back`. Check my logic, but I think we could collapse things a bit by relying on the fact that these are checked in order, so `[_, _, :f, :f, _, _]` would suffice as long as it came after `:front` and `:back`. Similarly, the `:middle` cases could be handled with just `[:f, _, _, _, _, :f]` as the last check (or just use a default).

panentheos (approved):
See comment, then looks good!

lib/content/message/predictions.ex
PaulJKim: Ah good catch. I meant to add a default to catch the `:front_and_back` case but must've skipped it. I think you're right though about collapsing. Will check the logic on that and update the PR and merge if all looks good

PR #678, opened 2023-08-24
OL crowding read loop selection

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🟠🧑‍🤝‍🧑🔊 Interrupt Read Loop Selection Logic](https://app.asana.com/0/1185117109217413/1205194882142438/f)

This PR adds logic for whether crowding data should be read out with an "Approaching" announcement, an "Arriving" announcement" or not at all.


panentheos (approved):
Looks good!

PR #682, opened 2023-08-30
OL crowding message creation

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🟠🧑‍🤝‍🧑🔊 Audio Message Creation](https://app.asana.com/0/1185117109217413/1205194882142440/f)

This PR adds new audio for attaching crowding info to Approaching and Arriving messages.

List of new take ids from ARINC:
- TAKE_ID 870 = near front of train
- TAKE_ID 871 = near back of train
- TAKE_ID 872 = middle of train
- TAKE_ID 873 = front and back of train
- TAKE_ID 874 = not crowded
- TAKE_ID 875 = somewhat crowded
- TAKE_ID 876 = crowded


lib/content/audio/approaching.ex
```diff
@@ -41,8 +45,13 @@ defmodule Content.Audio.Approaching do
...
           end
 
-        var ->
+        {var, nil} ->
           {:canned, {"103", [var], :audio_visual}}
+
+        {var, crowding_description} ->
+          {:canned,
+           {"104", [var, Content.Utilities.crowding_description_var(crowding_description)],
+            :audio_visual}}
```
panentheos, line 54: Since this is visual as well, presumably the text will be concatenated together. Do we need a space?

lib/content/utilities.ex
```diff
@@ -110,4 +110,17 @@ defmodule Content.Utilities do
...
+  def crowding_description_var(crowding_description) do
+    case crowding_description do
+      {:front, _} -> "870"
+      {:back, _} -> "871"
+      {:middle, _} -> "872"
+      {:front_and_back, _} -> "873"
+      {:train_level, :not_crowded} -> "874"
+      {:train_level, :some_crowding} -> "875"
+      {:train_level, :crowded} -> "876"
+      _ -> nil
```
panentheos, line 123: We should probably make this a "space" var so it's valid, since we don't have any special error checking for this case.

panentheos (approved):
See comments, then looks good!

lib/content/audio/approaching.ex
PaulJKim: good question. I don't know for sure and would have to ask Wabtec about that, but we should probably include one for good measure

PR #685, opened 2023-08-31
Revert "Revert "Determine emptier car location and crowding description""

PaulJKim:
Reverts mbta/realtime_signs#684

Errors were occurring in Dev because the `multi_carriage_details` field is still only in dev-green and that was causing `get_crowding_description` to attempt to call `Enum.map(nil)`.

Adding a default of `[]` to the parsing logic. Tested against the prod RTR feed to make sure there are no errors.

### DIFF
--- a/lib/engine/locations.ex
+++ b/lib/engine/locations.ex
@@ -142,11 +142,11 @@ defmodule Engine.Locations do
       trip_id: location["vehicle"]["trip"]["trip_id"],
       consist: location["vehicle"]["vehicle"]["consist"],
       multi_carriage_details:
-        location["vehicle"]["multi_carriage_details"] &&
-          Enum.map(
-            location["vehicle"]["multi_carriage_details"],
-            &parse_carriage_details/1
-          )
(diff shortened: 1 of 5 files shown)


lib/engine/locations.ex
```diff
@@ -142,11 +142,11 @@ defmodule Engine.Locations do
...
       consist: location["vehicle"]["vehicle"]["consist"],
       multi_carriage_details:
-        location["vehicle"]["multi_carriage_details"] &&
-          Enum.map(
-            location["vehicle"]["multi_carriage_details"],
-            &parse_carriage_details/1
-          )
+        (location["vehicle"]["multi_carriage_details"] &&
+           Enum.map(
+             location["vehicle"]["multi_carriage_details"],
```
panentheos, line 147: nit: If we do `location["vehicle"]["multi_carriage_details"] || []`, we could remove the outer guard and map over it unconditionally, which might be a little cleaner.

panentheos (approved):
See comment, then looks good.

PR #689, opened 2023-09-11
Add more detailed audio logging

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Improve send_audio logging in Splunk](https://app.asana.com/0/1185117109217413/1205407137195560/f)

Include the type of audio struct that a message is being derived from in our `send_audio` and `send_custom` audio logs.

Also include a string representation of any fields and their values that belong to the struct.

### DIFF
--- a/lib/pa_ess/http_updater.ex
+++ b/lib/pa_ess/http_updater.ex
@@ -124,7 +124,15 @@ defmodule PaEss.HttpUpdater do
           t()
         ) ::
           post_result()
-  defp process_send_audio(station, zones, audio, priority, timeout, sign_id, state) do
+  defp process_send_audio(
+         station,
+         zones,
+         %{__struct__: struct} = audio,
+         priority,
+         timeout,
+         sign_id,
+         state
+       ) do
     case Content.Audio.to_params(audio) do
       {:canned, {message_id, vars, type}} ->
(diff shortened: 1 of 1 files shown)


lib/pa_ess/http_updater.ex
```diff
@@ -141,7 +149,15 @@ defmodule PaEss.HttpUpdater do
...
 
         {arinc_ms, signs_ui_ms, result} = send_payload(encoded, state)
-        log("send_audio", encoded, arinc_ms: arinc_ms, signs_ui_ms: signs_ui_ms, sign_id: sign_id)
+
+        log("send_audio", encoded,
+          arinc_ms: arinc_ms,
+          signs_ui_ms: signs_ui_ms,
+          sign_id: sign_id,
+          message_type: to_string(struct) |> String.split(".") |> List.last(),
+          message_details: Map.from_struct(audio) |> inspect()
```
panentheos, line 158: Thinking out loud: Logging the whole struct is probably ok for now, but we should keep in mind that the refactoring effort is aiming to substantially change or remove many of these structs, so we shouldn't build any dashboards or alerting based on their internal contents.

panentheos (approved):
Looks good, see comment.

PR #690, opened 2023-09-12
replace 20+ minutes with actual predictions

panentheos:
#### Summary of changes
**Asana Ticket:** [Replace 20+ min with actual predictions](https://app.asana.com/0/1185117109217413/1203830107648831/f)

See ticket for complete description. Additional notes:
* Fixes an issue where we were inadvertently doing immediate audio announcements of 20+ predictions.
* Fixes an edge case where predictions with nil arrival or departure times would always suppress stopped status.

### DIFF
--- a/lib/signs/utilities/predictions.ex
+++ b/lib/signs/utilities/predictions.ex
@@ -10,6 +10,8 @@ defmodule Signs.Utilities.Predictions do
   alias Signs.Utilities.SourceConfig
 
   @reverse_prediction_certainty 360
+  @max_prediction_sec 60 * 60
+  @reverse_prediction_cutoff_sec 20 * 60
 
   @spec get_messages(Signs.Realtime.predictions(), Signs.Realtime.t()) ::
           Signs.Realtime.sign_messages()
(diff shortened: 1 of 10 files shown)


lib/signs/utilities/predictions.ex
```diff
@@ -150,19 +152,26 @@ defmodule Signs.Utilities.Predictions do
...
+  defp stopped_train?(%{
+         boarding_status: boarding_status,
+         seconds_until_arrival: seconds_until_arrival,
+         seconds_until_departure: seconds_until_departure,
+         arrival_certainty: arrival_certainty,
+         departure_certainty: departure_certainty
+       }) do
+    # Note: This performs a similar (but not identical) calculation to the one in the Message
+    # code for determining whether a prediction will show an approximate time. Ideally they
+    # should both call the same logic.
```
PaulJKim, line 169: Any reason to not go ahead and unify that logic now? I may be missing precisely how they're not identical. Could we do something like calling the logic in `Content.Message.Predictions.compute_minutes/2` but only take the second element in the result(s)?

PaulJKim (approved):
Minor questions/comments but otherwise looks good!

lib/signs/utilities/predictions.ex
panentheos: The complicating factor is that `compute_minutes` depends on selecting either the departure or the arrival time, which is currently embedded in `terminal` and `non_terminal`, and choosing between those functions requires looking up the source, e.g. with `get_source_by_stop_and_direction`. Pulling apart those layers would have added a lot of bulk to this PR.

PR #691, opened 2023-09-12
Add workflow for deploying to dev-green

PaulJKim:
#### Summary of changes
Adds a new GH workflow for deploying to Realtime Signs dev-green.

Note: The dev-green environment is not set up yet as of 9/12. Currently working with infra to do that.

### DIFF
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -1,0 +1,34 @@
+name: Deploy to Dev-green
+
+on:
+  workflow_dispatch:
+
+jobs:
+  Build:
+    runs-on: windows-2019
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


PaulJKim:
Question for @paulswartz - If I recall correctly, on-prem deploys to `prod` do not build a docker image and re-use the latest one that was deployed to `dev`. If we instituted a `dev-green` environment and this new GH workflow for deploying to it, could that cause `dev-green` images to potentially get deployed to `prod` if `dev-green` was the last environment deployed to by someone rather than `dev`?

paulswartz:
@PaulJKim the prod deploy still uses the SHA of the branch being deployed, so as long as you're only deploying `main` to prod, you should be fine.

panentheos (approved):
Looks good!

PR #693, opened 2023-09-14
Fix dev green deploys

PaulJKim:
#### Summary of changes
We need to bind the dev-green container to a different port as port 80 is already being used by dev. Tested the workflow from the branch and was able to successfully deploy the instance.

Upon doing this, I started seeing 403 errors in Splunk indicating that Realtime Signs dev-green wasn't able to read the config file from S3. I realized that the resource list for S3 needed to be updated in Terraform to allow Realtime Signs to read the file.

While making the Terraform update, I also decided to make a similar change for the `headend.json` file that we keep in S3. This file was added fairly recently to let us manage what headend server Realtime Signs sends messages through via S3. For dev-green, the IP for the headend server should be `nil` because there is no dev-green equivalent amongst the PAESS headend servers. When the config value is set to `nil`, Realtime Signs will skip sending messages to the headend server and only send them to Signs UI, which is desirable for dev-green.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -24,6 +24,7 @@ if config_env() != :test do
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
     message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
     monitoring_api_key: System.get_env("MONITORING_API_KEY")
+    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH", "/headend.json"),
 end
 
 if config_env() == :dev do
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -24,7 +24,7 @@ if config_env() != :test do
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
     message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
     monitoring_api_key: System.get_env("MONITORING_API_KEY"),
-    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH", "/headend.json"),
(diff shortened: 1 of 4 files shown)


lib/engine/config.ex
```diff
@@ -124,6 +124,10 @@ defmodule Engine.Config do
     updater = Application.get_env(:realtime_signs, :external_config_getter)
 
     case updater.get_active_headend_ip() do
+      {:ok, ""} ->
```
PaulJKim, line 127: The thought here is to add a new file to the `mbta-signs-dev` bucket but have the value for the headend IP as `""`

panentheos: Could we give the new environment its own separate bucket? That way we could keep everything isolated.

PaulJKim: I had planned to do it that way at first, but the infra team advised to reuse an existing bucket

config/runtime.exs
```diff
@@ -23,7 +23,8 @@ if config_env() != :test do
     message_log_s3_bucket: System.get_env("MESSAGE_LOG_S3_BUCKET"),
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
     message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
-    monitoring_api_key: System.get_env("MONITORING_API_KEY")
+    monitoring_api_key: System.get_env("MONITORING_API_KEY"),
+    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH", "/headend.json")
```
panentheos, line 27: Assuming this is set correctly in all environments, I think it would be better to avoid having a default value.

lib/engine/config.ex
panentheos: In that case, if we make the json value `null` instead of `""`, the existing code path should work without modifications.

panentheos (approved):
Looks good! (Deploy conditional on the new env var being set in prod, of course)

PR #700, opened 2023-09-22
convert crowding tests to top-level tests

panentheos:
#### Summary of changes
**Asana Ticket:** [RTS Refactor: Audio Logic Revision](https://app.asana.com/0/1185117109217413/1205450681869477/f)

This converts the crowding tests to top-level sign tests, as part of the audio overhaul. Notes:
* As of now, the actual crowding messages are still disabled, so the tests reflect the lack of crowding messages. I tested locally with the messages temporarily enabled, and the results look correct.
* The non-test changes parameterize the locations engine so that it can be mocked.
* A few other tests are now generating audio due to the default sign struct changing to use `announce_arriving?: true` (which better reflects platform signs).
* A couple of the crowding test cases were combined, but the overall coverage is the same.


PaulJKim (approved):
Looks good

PR #703, opened 2023-09-28
calculate all routes from signs json

panentheos:
#### Summary of changes

This is a bit of pre-work that came out of investigating the linked ticket. Calculates the list of route ids dynamically from the config, rather than hard-coding. This should ensure we're getting alerts for all bus routes, not just SL.


PaulJKim (approved):
Looks good!

PR #704, opened 2023-10-02
Single line alerts on the first line of a MZ sign should trigger a flip

PaulJKim:
#### Summary of changes
**Asana Ticket:** [[bug] Single-line alerts at Ashmont lobby on top line](https://app.asana.com/0/1185117109217413/1205634330387875/f)

Left out single-line alert messages when we added in the early AM logic and the `flip?` function. These messages take up 24 characters and therefore need to be on the bottom line of the sign.

Screenshot of fix running locally:
Screenshot 2023-10-02 at 5 01 31 PM


panentheos (approved):
Looks good.

PR #705, opened 2023-10-03
restructure bus configs to support route alerts

panentheos:
#### Summary of changes

This changes the structure of bus config data so that it will be straightforward to add support for showing alerts for individual routes. This PR should _not_ change the behavior of the system.

Adds a new level of object to the JSON, called configs. Each config represents a distinct logical grouping of routes that the sign reports on. Configs contain sources, which describe which predictions (and in the future, alerts) to match for that config. Most configs have a single source, except for the signs where we aggregate multiple routes into a single message, e.g. the SL waterfront. This approach turns these aggregations into first-class concepts, so that the alert code will be able to respect them.

### DIFF
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -15,10 +15,10 @@ defmodule Signs.Bus do
     :text_zone,
     :audio_zones,
     :max_minutes,
-    :sources,
-    :top_sources,
-    :bottom_sources,
-    :extra_audio_sources,
+    :configs,
+    :top_configs,
+    :bottom_configs,
+    :extra_audio_configs,
     :chelsea_bridge,
     :read_loop_interval,
     :read_loop_offset,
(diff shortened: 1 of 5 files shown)


lib/signs/bus.ex
```diff
@@ -407,6 +415,21 @@ defmodule Signs.Bus do
...
+
+  defp configs_content(configs, predictions_lookup) do
+    Enum.flat_map(configs, fn config ->
+      Stream.flat_map(config.sources, fn source ->
+        Map.get(predictions_lookup, {source.stop_id, source.route_id, source.direction_id}, [])
+      end)
+      |> Enum.group_by(&PaEss.Utilities.headsign_key(&1.headsign))
+      |> Enum.map(fn {_, list} ->
+        {:predictions, Enum.sort_by(list, & &1.departure_time, DateTime)}
+      end)
```
panentheos, line 428: Note: This is where the actual alert message generation will happen. If we don't find any predictions for a config, we can consult the alert data and produce an `:alert` structure here if appropriate. We'll then need to sort them to the end, and write code for formatting them and reading them.

PaulJKim (approved):
Tested locally and looks good

PR #706, opened 2023-10-03
Manual runs of message log jobs as Tasks

PaulJKim:
These manual runs currently run synchronously and the message latency report job appears to time out after a while if manually triggered via an HTTP request.

Judging from the logs, it looks like an automatic retry is triggered around the 1 minute mark which is probably killing the original process and preventing the job to run to completion which typically takes about 2 minutes. Then an error is finally returned to the client around the 2 minute mark which seems to suggest a 60 second default timeout from Phoenix for any given request. This seems to check out based off a few forum posts online. Rather than manually overriding this and extending the timeout, we can kick off the job as a background task and respond to the client immediately.


panentheos (approved):
Looks good!

PR #707, opened 2023-10-10
convert audio readout tests to top-level tests

panentheos:
#### Summary of changes

This rewrites the remaining audio utility tests as top-level tests, to assist refactoring efforts. Note that some tests were not directly converted, since they were already covered by other cases. Also, the existing code exhibits some likely undesirable edge cases (due to implementation details of the previous announcement logic), which are called out as comments in the new tests. We should start a followup effort to simplify and standardize that behavior.


test/signs/realtime_test.exs
```diff
@@ -708,6 +708,295 @@ defmodule Signs.RealtimeTest do
...
+
+      expect_messages({"Ashmont        ARR", "Ashmont      2 min"})
+      expect_audios([{:canned, {"103", ["32107"], :audio_visual}}])
+      Signs.Realtime.handle_info(:run_loop, %{@sign | tick_read: 0, announced_arrivals: ["1"]})
+    end
+
+    test "only reads the bottom line when the bottom line is arriving on a multi_source sign for heavy rail" do
+      # Note: This case doesn't come up during normal operation, because non-terminal signs
+      # announce arriving trains, and that announcement delays readouts until after the train
+      # has passed. However, for consistency, we should read the full sign.
```
PaulJKim, line 875: Think I'm having some trouble following these comments. Are these notes saying that the case that is mentioned in the test description doesn't happen during normal operations? For this comment in particular, why does the fact that non-terminal signs announce arriving trains mean that we wouldn't announce arriving only on the bottom line?

panentheos: Yeah, some of these cases don't tend to come up, due to the way announcements delay readouts. But we still need to define the behavior, so we might as well make it consistent. In this case, if the bottom line is ARR, the readout will _skip_ reading the top line (but only for heavy rail mezzanine signs). That seems like odd behavior, hence the recommendation.

PaulJKim: I definitely think the heavy rail only distinction is a little odd, but I'm not 100% sure if we want to do away with prioritizing the readout of the second line when it's on ARR in mezzanine zones. I believe the intention of the past designers/engineers was to announce that information to riders asap since at the point when a train is arriving, even a few seconds of a delay could make the difference between someone just making the train and missing it.

PaulJKim: That said, I agree the full sign should probably be read, just while still prioritizing ARR info if possible. I suppose if they're both arriving at the same time though, we'd be forced to delay one or the other.

panentheos: The new (separate) announcement code handles this prioritization. As soon as either line becomes ARR, it will trigger the immediate announcement for just that arrival, which will take priority over passive readouts. This code is (now) just handling an unlikely case where a train somehow remains ARR for a full 2 minutes after the initial announcement, and the normal read loop is triggered. I think we should default to an "audio equivalence" approach in that case, and just read the whole sign.

panentheos: I guess we could re-introduce some basic sorting of the audio messages, on the off chance this does happen.

PaulJKim (approved):

PR #708, opened 2023-10-11
display individual bus route alerts

panentheos:
#### Summary of changes

This adds "no service" messages for individual bus routes when there is an alert present, and no predictions. Also upgrades the stop-level "no service" message with new audio, and fixes a bug in the config loading that prevented bus alerts from loading.

Includes a mapping from bus route to destination name, which was pulled from the API using the included one-off script.


PaulJKim (approved):
Code looks good. Were you able to test this locally though? I'm struggling to create an alert that will trigger the alert messaging, but I'm also not familiar with how OIO's set up bus alerts of this sort.

panentheos:
I did some local testing, but only with synthetic alert statuses, not producing the status with an actual alert. I had assumed that the data flow would work the same as it currently does for rail alerts, but I agree it would be good to do some end-to-end verification first.

panentheos:
Ok, I tested locally by adding a route alert in dev (and locally blocking predictions for that route), and the alert message showed as expected.

PR #709, opened 2023-10-13
OL crowding revision

PaulJKim:
#### Summary of changes
**Asana Ticket:** [🟠🧑‍🤝‍🧑🔊 Text/Audio revisions](https://app.asana.com/0/1185117109217413/1205551289750420/f)

This PR makes some minor tweaks to the OL crowding logic.
- Crowding levels are now based off of occupancy percentages rather than occupancy status
- "some crowding" and "not crowded" train-level readouts are no longer used

### DIFF
--- a/lib/content/message/predictions.ex
+++ b/lib/content/message/predictions.ex
@@ -206,9 +206,10 @@ defmodule Content.Message.Predictions do
           else: :f
       end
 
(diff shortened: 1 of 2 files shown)


panentheos (approved):
Looks good.

PR #711, opened 2023-10-19
prevent count-up by prediction

panentheos:
#### Summary of changes
**Asana Ticket:** [RTS Refactor: Count-Up Suppression](https://app.asana.com/0/1185117109217413/1205499733465428/f)

This performs count-up suppression at the prediction level, instead of at the message level, mirroring the approach taken in the bus code. This results in fewer special cases, and should catch a few edge cases that the old code would miss.


PaulJKim (approved):
LGTM!

PR #712, opened 2023-10-19
Limit headways at end of service

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Limit Headway display during end of service](https://app.asana.com/0/1205101743911491/1205695443333800/f)

Change to stop showing headway messages slightly before the last scheduled trip to be a little more accurate. 

Also added a 1 min buffer to both the start of headway messages as well as the end to be slightly more accurate. For example, if the last scheduled train is at 1:00AM and the low range of headways is 5 minutes, then it would still be fair to show headways until the clock changes to 12:56AM, but 12:56AM would be the time of the last scheduled train - **4** minutes, rather than 5 minutes.

### DIFF
--- a/lib/engine/scheduled_headways.ex
+++ b/lib/engine/scheduled_headways.ex
@@ -8,6 +8,7 @@ defmodule Engine.ScheduledHeadways do
   use GenServer
   require Logger
   require Signs.Utilities.SignsConfig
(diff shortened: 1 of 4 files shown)


panentheos (approved):
Looks good!

PR #713, opened 2023-10-20
remove unused headway message code

panentheos:
#### Summary of changes

This removes various headway-related code that is unused or unreachable:
* Calculation of dynamic headway ranges from schedule data
* Display of previous departures
* Display of first departures
* Display of headways "up to" a value
* `:none` as a potential headway value


PaulJKim (approved):
Looks good. Thanks for cleaning all this up!

PR #714, opened 2023-10-20
Fix GL routes in signs.json

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Investigate modifying headway destination based on (GL branch) service suspensions](https://app.asana.com/0/1185117109217413/1205572772013992/f)

In #663 we removed an old piece of logic that filtered predictions based off of the `routes` field in the source config. Now that that's gone, we don't need to do weird stuff in the `signs.json` file in order to get outlier predictions to show (ex. the occasional D train going down the Medford branch).

This change adjusts the routes fields for a handful of GL stations so that they accurately reflect what branches typically serve them. The change is also meant to set us up in anticipation of future logic for checking for downstream suspensions.

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -6483,7 +6483,6 @@
             "stop_id": "70150",
             "direction_id": 1,
             "routes": [
-              "Green-B",
               "Green-C",
               "Green-D"
             ],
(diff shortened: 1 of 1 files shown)


panentheos (approved):
Looks good.

PR #716, opened 2023-10-23
fetch routes destinations from API

panentheos:
#### Summary of changes
**Asana Ticket:** [Add dynamic API calls for individual bus route alerts](https://app.asana.com/0/1185117109217413/1205660768174999/f)

This adds a new engine to dynamically pull bus route destinations from the API. It polls for new information every 30 seconds, and stores the result in ETS for quick lookups. In the unlikely event that a lookup fails (e.g. due to the API being down on startup), the downstream content generation will fall back on blank values.

### DIFF
--- a/lib/engine/routes.ex
+++ b/lib/engine/routes.ex
@@ -1,0 +1,61 @@
+defmodule Engine.Routes do
+  use GenServer
+  require Logger
+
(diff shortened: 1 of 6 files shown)


PaulJKim (approved):
Looks good!

PR #719, opened 2023-10-26
Simplify passive readouts

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Simplify sign readout behavior](https://app.asana.com/0/1185117109217413/1205674914175518/f)

Simplifies passive sign readout.

Slight change in behavior to read out both lines of sign when one of the lines is arriving.


panentheos (commented):
This is great! See a few comments on implementation details, but the approach looks spot on.

lib/signs/utilities/audio.ex
```diff
@@ -11,6 +11,118 @@ defmodule Signs.Utilities.Audio do
...
+         {%Message.Predictions{minutes: minutes, route_id: route_id} = prediction, index}
+       ) do
+    case minutes do
+      :boarding ->
+        Audio.TrainIsBoarding.from_message(prediction)
+
+      :arriving when route_id in @heavy_rail_routes ->
+        Audio.TrainIsArriving.from_message(
+          prediction,
+          prediction.crowding_data_confidence == :high
```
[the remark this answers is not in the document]
PaulJKim: ah true, good point!

lib/signs/utilities/audio.ex
```diff
@@ -11,6 +11,108 @@ defmodule Signs.Utilities.Audio do
...
+
+      minutes when is_integer(minutes) ->
+        Audio.NextTrainCountdown.from_message(prediction)
+
+      _ ->
+        []
+    end
+  end
+
+  defp get_prediction_readout({%Message.StoppedTrain{} = prediction, _}) do
```
panentheos, line 112: This signature looks incorrect, should probably be:
```suggestion
  defp get_prediction_readout(%Message.StoppedTrain{} = prediction) do
```

panentheos (approved):
See comment, then looks good.

lib/signs/utilities/audio.ex
PaulJKim: woops yes, good catch

PR #723, opened 2023-11-07
Split Lechmere into two visual zones

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Split Lechmere busway into two visual zones](https://app.asana.com/0/1185117109217413/1205857754053049/f)

Split the Lechmere busway into two visual zones that share the same audio zone.

Inner zone should show routes 69 and 80 and the outer zone should show route 87 and 88.

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -9152,7 +9152,7 @@
     "max_minutes": 60
   },
   {
-    "id": "bus.Lechmere_bus_mezzanine",
+    "id": "bus.Lechmere_inner",
     "pa_ess_loc": "SLEC",
     "read_loop_interval": 360,
     "read_loop_offset": 180,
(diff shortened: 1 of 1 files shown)


priv/signs.json
```diff
@@ -9179,7 +9179,21 @@
...
+    "max_minutes": 60
+  },
+  {
+    "id": "bus.Lechmere_outer",
+    "pa_ess_loc": "SLEC",
+    "read_loop_interval": 360,
+    "read_loop_offset": 0,
+    "text_zone": "c",
+    "audio_zones": [
+      "m"
```
panentheos, line 9193: This configuration will perform two separate readouts on the same zone, offset by a couple minutes. I thought we were going to handle this similar to Nubian, where the visual messages are split but there's a single combined audio readout via the `extra_audio_configs` mechanism, so people don't get confused if they don't hear their bus.

PaulJKim: Ah right, I forgot we added that feature to the bus signs. Just adjusted to the config

panentheos (approved):
Looks good!

PR #725, opened 2023-11-22
Migrate off of consist field

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Transition from 'consist' to `multi_carriage_details'](https://app.asana.com/0/1185117109217413/1205822469391764/f)

The transit data team is moving away from the `consist` field by mid December. We only use consist data to identify when a Red Line train so the migration doesn't have a major impact on us, but does mean that we need to fetch the same info from the locations feed now rather than directly from the predictions feed.

### DIFF
--- a/lib/content/message/predictions.ex
+++ b/lib/content/message/predictions.ex
@@ -103,7 +103,7 @@ defmodule Content.Message.Predictions do
           trip_id: prediction.trip_id,
           direction_id: prediction.direction_id,
           width: width,
-          new_cars?: prediction.new_cars?,
+          new_cars?: sign.location_engine.for_vehicle(prediction.vehicle_id) |> new_cars?(),
           station_code: station_code,
           zone: zone,
           platform: platform,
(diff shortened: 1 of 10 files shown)


lib/content/message/predictions.ex
```diff
@@ -260,6 +266,27 @@ defmodule Content.Message.Predictions do
...
+         route_id: route_id
+       }) do
+    Enum.any?(multi_carriage_details, fn carriage ->
+      # See http://roster.transithistory.org/ for numbers of new cars
+      case Integer.parse(carriage.label) do
+        :error ->
+          false
+
+        {n, _remaining} ->
+          route_id == "Red" and 1900 <= n and n <= 2151
```
panentheos, line 285: Are we not replicating the orange line case here?

panentheos (approved):
See question, then looks good.

lib/content/message/predictions.ex
PaulJKim: We actually stopped making the new orange line cars announcements a while back but in that change, I just removed the reference to that message ID that was being used in the `Approaching` struct rather than removing the logic in the `consist_is_new?` function. #526 for reference

PR #726, opened 2023-12-08
Enable crowding audio at Back Bay platform for 12-11 demo

PaulJKim:
Ad-Hoc

To be deployed around 8am on Monday morning (12/11/23) for an in-station demo

### DIFF
--- a/lib/signs/utilities/audio.ex
+++ b/lib/signs/utilities/audio.ex
@@ -249,7 +249,7 @@ defmodule Signs.Utilities.Audio do
     new_audios =
       Enum.map(new_audios, fn %{__struct__: audio_type} = audio ->
         if audio_type in [Audio.Approaching, Audio.TrainIsArriving] and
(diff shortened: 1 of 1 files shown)


lib/signs/utilities/audio.ex
```diff
@@ -248,7 +248,8 @@ defmodule Signs.Utilities.Audio do
     # Disable crowding messages for now
     new_audios =
       Enum.map(new_audios, fn %{__struct__: audio_type} = audio ->
-        if audio_type in [Audio.Approaching, Audio.TrainIsArriving] and sign.id not in [] do
+        if audio_type in [Audio.Approaching, Audio.TrainIsArriving] and
+             sign.id not in ["back_bay_northbound", "back_back_southbound"] do
```
panentheos, line 252: There's a typo in the second string.

panentheos (approved):
See comment, then looks good.

PR #727, opened 2023-12-12
Default to arrival time for schedules

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Revise display logic for departure/arrival times](https://app.asana.com/0/1185117109217413/1206103644850840/f)

Small change to add some resiliency to our schedules logic considering some weird data we saw from the API.


panentheos (approved):
Seems good to me.

PR #729, opened 2023-12-18
Union Sq defaults to using custom alert messaging

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Adjust the default message for alerts that include Union Sq](https://app.asana.com/0/1185117109217413/1206049304225903/f)

Union Sq never gets shuttle service so OIOs have requested that we change Union Sq's default alert messaging to say

```
No Train Service
Use Routes 86, 87, or 91
```

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -180,18 +180,15 @@ defmodule Signs.Utilities.Messages do
   end
 
   defp get_alert_messages(alert_status, %{text_id: {"GUNS", _}}) do
-    cond do
-      alert_status in [:none, :alert_along_route] ->
-        nil
-
-      true ->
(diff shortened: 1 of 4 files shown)


lib/signs/utilities/messages.ex
```diff
@@ -179,6 +179,16 @@ defmodule Signs.Utilities.Messages do
     Content.Message.Empty.new()
   end
 
+  defp get_alert_messages(alert_status, %{text_id: {"GUNS", _}}) do
+    cond do
```
panentheos, line 183: nit: for just two branches an `if` may be cleaner

panentheos (approved):
Code looks good, but it might be good to write a quick top-level test that captures this case to make sure the behavior is preserved through refactors.

PR #730, opened 2023-12-19
hotfix: treat unknown crowding as `:train_level`

paulswartz:
#### Summary of changes
See below

### DIFF
--- a/lib/content/message/predictions.ex
+++ b/lib/content/message/predictions.ex
@@ -197,9 +197,7 @@ defmodule Content.Message.Predictions do
 
   defp calculate_crowding_data_confidence(_prediction, _location), do: nil
 
-  defp get_crowding_description([]), do: nil
-
-  defp get_crowding_description(carriage_details) do
+  defp get_crowding_description(carriage_details) when length(carriage_details) == 6 do
     crowding_levels =
       Enum.map(carriage_details, &occupancy_percentage_to_crowding_level(&1.occupancy_percentage))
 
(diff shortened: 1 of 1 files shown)


panentheos:
This guards against trying to process a `multi_carriage_details` with < 6 elements, which caused a couple crashes. The underlying data issue is also being addressed, but it may take some time, and even afterward it seems prudent to avoid crashing in case something similar happens again.

Note: The first commit on this branch was deployed to production as a hotfix this morning. I added the second commit as a more targeted fix for the specific case. Technically that makes the original fix redundant, but I'm inclined to leave it as a reasonable fallback for the "emptier location" logic.

lib/content/message/predictions.ex
```diff
@@ -197,9 +197,7 @@ defmodule Content.Message.Predictions do
 
   defp calculate_crowding_data_confidence(_prediction, _location), do: nil
 
-  defp get_crowding_description([]), do: nil
-
-  defp get_crowding_description(carriage_details) do
+  defp get_crowding_description(carriage_details) when length(carriage_details) == 6 do
```
paulswartz, line 200: suggestion(non_blocking): it likely doesn't matter in this case, but in general it's good to avoid calculating the full length if you don't need it. Here, you could replace it with:
```suggestion
  defp get_crowding_description([_, _, _, _, _, _]) do
```
For more, see https://www.erlang.org/docs/26/efficiency_guide/commoncaveats.html#length-1

panentheos: The example in the docs works because it can stop after the first 3 elements, but I'd be surprised if this code was able to somehow complete the pattern match without effectively traversing the whole list. Still, I like how the match looks visually in this code. Thanks!

paulswartz: This pattern match works the same way: it can stop after 6 elements, where the length() call would need to continue.

PaulJKim (approved):

PR #732, opened 2024-01-08
Update dockerfile and actions

PaulJKim:
#### Summary of changes
**Asana Ticket:** [migrate realtime-signs to Linux cluster](https://app.asana.com/0/1185117109217413/1205187491334062/f)

New Dockerfile and github actions for moving Realtime Signs to run on linux

### DIFF
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -16,7 +16,7 @@ jobs:
     concurrency: dev-green-linux
     env:
       ECS_CLUSTER: linux-staging
-      ECS_SERVICE: realtime-signs-dev-green
+      ECS_SERVICE: realtime_signs-dev-green
     steps:
       - uses: actions/checkout@v3
       - uses: mbta/actions/build-push-ecr@v2
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -9,6 +9,9 @@ jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read    
     environment: dev-green-linux
     concurrency: dev-green-linux
     env:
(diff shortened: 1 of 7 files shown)


.github/workflows/dev-green.yml
```diff
@@ -1,35 +1,37 @@
...
-      - name: Checkout repository
-        uses: actions/checkout@v2
-      - uses: mbta/actions/build-push-ecr@v2
+      - uses: actions/checkout@v3
+      - uses: mbta/actions/build-push-ecr@v1
         id: build-push
         with:
-          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
PaulJKim, line 23: Realtime Signs doesn't have these secrets stored in the repo environments so I guess we'll have to add them. I thought though that we wanted to avoid saving AWS credentials in GitHub repos directly, or has that changed?

paulswartz: You should be using `v2` which uses a different connection type to AWS: https://github.com/mbta/busloc/blob/master/.github/workflows/staging.yml#L25-L34

Dockerfile
```diff
@@ -1,51 +1,52 @@
...
 
 RUN mix deps.compile
 
 COPY lib lib
 COPY priv priv
 
-COPY config/runtime.exs config\\runtime.exs
-RUN mix release
+COPY config/runtime.exs config/runtime.exs
+RUN mix release linux
```
paulswartz, line 29: I got an error at this line; did you mean to also check in a change to `mix.exs`?

paulswartz (commented):
You might want to include a `docker.yml` to ensure that the Dockerfile builds: https://github.com/mbta/busloc/blob/master/.github/workflows/docker.yml

Dockerfile
PaulJKim: Ah yes, thank you

paulswartz (approved):

PR #734, opened 2024-01-29
allow application to continue running despite sign process crashes

panentheos:
#### Summary of changes
**Asana Ticket:** [Allow partial failures in RTS](https://app.asana.com/0/1185117109217413/1206301659818088/f)

This changes the initial delay on sign processes from 1 second to 5 seconds, which means repeated sign crashes no longer cause the app to restart. In addition:
* Adds human-readable names to all processes, for more useful logging.
* Reworks `Engine.Predictions` to store non-revenue vehicle state in ETS, removing a bottleneck that was causing intermittent crashes of `HeadwayAnalysis` processes.
* Cleans up and standardizes some supervisor initialization code.


PaulJKim (approved):
Looks good to me

PR #746, opened 2024-04-11
Last Trip of the Day phase 1

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Read in the "last train" information from transit data feed and display "end of service" information](https://app.asana.com/0/1185117109217413/1206605275887005/f)

This PR adds new state for tracking recent departures within the last hour as well as any trips that were marked as a last trip within the last hour.

It also adds logic for showing and playing visual content and audio indicating that service has ended for the night. The logic varies based on whether the source is for a single direction or two directions (platform vs mezzanine, for the most part). The logic is outlined in the Asana task.


lib/signs/utilities/messages.ex
```diff
@@ -122,6 +130,69 @@ defmodule Signs.Utilities.Messages do
...
+         {has_top_service_ended?, has_bottom_service_ended?},
+         {top_source, bottom_source}
+       ) do
+    cond do
+      has_top_service_ended? and has_bottom_service_ended? ->
+        {%Content.Message.LastTrip.StationClosed{}, %Content.Message.LastTrip.ServiceEnded{}}
+
+      has_top_service_ended? ->
+        if get_message_length(bottom_message) <= 18 do
+          {bottom_message,
```
panentheos, line 144: This works fine for typical mezzanine modes, but it doesn't handle cases where the lines no longer correspond 1-1 with the source configs, like combined headways, combined alerts, or JFK special paging. I think we'll need to take a similar approach to the early AM logic, and unpack those cases so we can handle them more holistically.

lib/engine/last_trip.ex
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  @impl true
+  def handle_info(:update, %{last_modified: last_modified} = state) do
+    schedule_update(self())
+
+    http_client = Application.get_env(:realtime_signs, :http_client)
+
+    new_last_modified =
+      case http_client.get(
```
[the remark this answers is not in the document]
PaulJKim: Yeah I definitely thought about that as well and that makes sense. My main concern was just that I didn't want to convolute the existing predictions engine code, but I think if we just encapsulate this logic here in its own module it should be just fine

lib/engine/last_trip.ex
```diff
@@ -0,0 +1,177 @@
...
+
+  defp insert_recent_departures(predictions_by_trip, state) do
+    current_time = Timex.now()
+
+    for {trip_id, predictions} <- predictions_by_trip,
+        prediction <- predictions,
+        prediction["departure"] do
+      seconds_until_departure = prediction["departure"]["time"] - DateTime.to_unix(current_time)
+
+      if seconds_until_departure <= 0 and abs(seconds_until_departure) <= @hour_in_seconds do
```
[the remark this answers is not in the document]
PaulJKim: Oh that's clever.. considering `@hour_in_seconds` is 3600 though, would this not end up being fairly inefficient of a check? I'm not exactly sure, but I'd assume `in` has a linear time complexity

lib/engine/last_trip.ex
```diff
@@ -0,0 +1,177 @@
...
+    end
+  end
+
+  defp clean_last_trips(state) do
+    :ets.tab2list(state.last_trips)
+    |> Enum.each(fn {trip_id, timestamp} ->
+      if Timex.diff(Timex.now(), timestamp, :seconds) > @hour_in_seconds do
+        :ets.delete(state.last_trips, trip_id)
+      else
+        :ok
```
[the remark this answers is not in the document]
PaulJKim: woops nope don't think so

lib/signs/realtime.ex
```diff
@@ -196,6 +211,25 @@ defmodule Signs.Realtime do
     {:noreply, state}
   end
 
+  defp has_service_ended_for_source?(sign, source, current_time) do
+    SourceConfig.sign_stop_ids(source)
+    |> Enum.all?(&has_last_trip_departed_stop?(&1, sign, current_time))
```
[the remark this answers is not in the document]
PaulJKim: Ah that's a good call out, it would be just one stop id and not all. And good question about the multiple branches too, I'll ask the transit data team about that one.

lib/signs/utilities/messages.ex
```diff
@@ -122,6 +130,69 @@ defmodule Signs.Utilities.Messages do
...
   end
 
+  defp get_last_trip_messages(
+         {top_message, bottom_message} = messages,
+         {has_top_service_ended?, has_bottom_service_ended?},
+         {top_source, bottom_source}
+       ) do
+    cond do
+      has_top_service_ended? and has_bottom_service_ended? ->
+        {%Content.Message.LastTrip.StationClosed{}, %Content.Message.LastTrip.ServiceEnded{}}
```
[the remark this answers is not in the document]
PaulJKim: Hmm right, we should probably show a prediction if there is one. We can add a check that there are no predictions first, before we get to this logic.

lib/engine/last_trip.ex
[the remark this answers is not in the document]
panentheos: According to the docs, `in` calls `Enum.member?`, which `Range` explicitly calls out as being efficient to use, so it should be okay (assuming the readability is okay)

lib/signs/realtime.ex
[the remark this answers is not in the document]
PaulJKim: So about routes, I'm not sure if it's something we need/want to handle right now as GL last trips are not going to be supported until last trips are a feature in Glides. It does sound like once it's implemented though, that it will be a last trip for every route, which means that we would probably need to key recent departures on stop_id + route, instead of just stop_id. It's not much more effort to include the route right now though so we can also just go ahead and do it to prepare for the future.

The routes discussion brings up an interesting case though. Since the red line has branching, there will be two last trips in both directions, but there aren't GTFS route_ids for the two red line branches so somehow the code will need to know to wait for two last trips at RL trunk stops, and just one for RL branch stops..

lib/signs/utilities/messages.ex
PaulJKim: I think at this point in the logic, the only case where the lines don't correspond 1-1 anymore is the JFK Mezzanine case where we have a NB platform prediction alongside SB headways. We could account for that specific case and unpack them into a regular prediction struct and paging headway struct respectively. I think at that point, the rest of the logic here would work. What are your thoughts on that or do you see a bug in that logic that I might've missed?

panentheos: One case I ran into was if a mezzanine sign comes up with no predictions, it can display a two-line combined headway message, which can then get sliced in half if one direction is LTOD. I think a similar thing can happen with alerts, though I didn't check that definitively.

PaulJKim: Ah right the headways is one that doesn't work. I wonder if the alerts case is something we can just pass along though considering that if in any case a MZ sign is showing a two-lined alert message, I believe that means service is definitely not running at that station and saying "No service use shuttle bus" or just simply "No service" essentially means the same thing as something like "Station closed service has ended for the night". Might be something to run by a PM/Designer.

PaulJKim: Spoke with Dan and we decided that for two-line combined alerts, we should show LTOTD message once service has fully ended. Otherwise, just show the alert.

lib/signs/utilities/last_trip.ex
```diff
@@ -0,0 +1,106 @@
...
+                 page?: false
+               }, unpacked_mz_top}
+            end
+
+          true ->
+            messages
+        end
+
+      has_service_ended? ->
+        if has_service_ended?,
```
panentheos, line 64: We should add a check for predictions here, to avoid suppressing them, right?

lib/signs/utilities/last_trip.ex
```diff
@@ -0,0 +1,106 @@
...
+    match?(%Content.Message.Predictions{}, message)
+  end
+
+  defp get_message_length(message) do
+    message_string = Content.Message.to_string(message)
+
+    if is_list(message_string) do
+      Enum.max_by(message_string, fn {string, _} -> String.length(string) end)
+      |> elem(0)
+      |> String.length()
```
panentheos, line 101: nit: This could be
```suggestion
      Stream.map(message_string, fn {string, _} -> String.length(string) end)
      |> Enum.max()
```

lib/engine/last_trip.ex
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  @impl true
+  def handle_cast(
+        {:update_recent_departures, new_recent_departures},
+        %{recent_departures: recent_departures_table} = state
+      ) do
+    current_recent_departures =
+      :ets.tab2list(recent_departures_table)
+      |> Stream.map(&{elem(&1, 0), elem(&1, 1)})
```
panentheos, line 81: Since these are already 2-tuples, this line looks like a no-op.

lib/content/audio/service_ended.ex
```diff
@@ -0,0 +1,96 @@
...
+      end
+    end
+
+    def to_tts(%Content.Audio.ServiceEnded{location: :station}) do
+      "This station is closed. Service has ended for the night."
+    end
+
+    def to_tts(%Content.Audio.ServiceEnded{location: :platform, destination: destination}) do
+      destination_string =
+        case(Utilities.destination_to_ad_hoc_string(destination)) do
```
panentheos, line 63: It should be safe to use `destination_to_ad_hoc_string` here and below without testing for failure or worrying about empty strings, since it has a valid value for all destinations.

panentheos (approved):
Overall this looks good. All the comments below are small things, and the only one that affects visible behavior is the one about not suppressing predictions in the platform case.

I spot-checked a few cases locally, but did not do extensive local testing.

PR #747, opened 2024-04-17
add update queues for SCUs

panentheos:
#### Summary of changes
**Asana Ticket:** [RTS: Support communication with new SCU app](https://app.asana.com/0/1185117109217413/1206141631184125/f)

Adds new processes for communicating with SCUs. Each SCU gets its own independent queue, using GenStage to simplify queueing and message passing. Expects to receive a new environment variable containing the mapping from SCU id to ip address, which will be provided separately.

Note that while this does add the processes themselves, they don't currently do anything because nothing is trying to call out to the new SCU app.


lib/pa_ess/scu_updater.ex
```diff
@@ -0,0 +1,92 @@
...
+    sign_ui_api_key = Application.get_env(:realtime_signs, :sign_ui_api_key)
+
+    if sign_ui_url do
+      http_poster.post(
+        "https://#{sign_ui_url}#{path}",
+        body,
+        [
+          {"Content-type", "application/json"},
+          {"x-api-key", sign_ui_api_key},
+          {"x-scu-id", scu_id}
```
PaulJKim, line 76: Just curious, why do we need to send the scu_id to Signs UI?

panentheos: This endpoint doesn't exist yet, but Signs UI will need this in order to attribute the data to the right sign, since the payload itself is not sufficient to differentiate it globally.

PaulJKim (approved):

PR #751, opened 2024-05-10
support communication with both SCU types

panentheos:
#### Summary of changes
**Asana Ticket:** [RTS: Support communication with new SCU app](https://app.asana.com/0/1185117109217413/1206141631184125/f)

This refactors the message sending code path to allow for branching based on whether the target SCU has been migrated.
* Reads the new `scus_migrated` field from the config, and makes it available via the engine
* Normalizes the subset of state related to station codes, zones, etc between sign types, so they can share the same helper functions
* Changes `PaEss.Updater` to a concrete new module, with a slightly more generic API
* Switches behavior based on the current migration status for the SCU. Note that the current code just logs an error on the new path, but this will be replaced with a call to the new `ScuQueue` in another PR.


cmaddox5 (approved):
Makes sense to me. Just to make sure I understand how this all connects, SignsUI will update `config.json` with the list of migrated SCUs, and that is what RTS puts into the new ETS table?

panentheos:
Yes, that's right.

PR #752, opened 2024-05-10
Don't page no service message on MZ signs

PaulJKim:
#### Summary of changes
Ad-hoc

Quick fix to prevent double-line paging on MZ signs during LTOTD treatment. This type of double paging causes a bug to occur in the vendor's software which leads to a crash and the signs going blank for some time.


cmaddox5 (approved):
Makes sense to me.

PR #754, opened 2024-05-20
fix custom mode not overriding last trip message

panentheos:
#### Summary of changes
This fixes a bug that was preventing custom mode from taking priority over LTOTD messages.


PaulJKim (approved):

digitalcora:
Since this bug slipped through the current test suite, can we add a test that would have caught it, to prevent regression and give us better coverage here? (If the fix is urgent and we're sure it works as intended, it could be a separate PR)

PR #755, opened 2024-05-28
add visual equivalent for rail TTS audio

panentheos:
#### Summary of changes

This refactors the rail audio modules to support optional visual text to go along with TTS announcements. The default approach is to just paginate the same text used for TTS, but this can be customized later on a per-message basis.

### DIFF
--- a/lib/content/audio.ex
+++ b/lib/content/audio.ex
@@ -19,6 +19,6 @@ defprotocol Content.Audio do
   @doc "Converts an audio struct to the mid/vars params for the PA system"
   @spec to_params(Content.Audio.t()) :: value()
   def to_params(audio)
-  @spec to_tts(Content.Audio.t()) :: String.t()
+  @spec to_tts(Content.Audio.t()) :: {String.t(), [{String.t(), String.t(), integer()}] | nil}
   def to_tts(audio)
 end
--- a/lib/pa_ess/utilities.ex
+++ b/lib/pa_ess/utilities.ex
@@ -697,4 +697,28 @@ defmodule PaEss.Utilities do
 
   def audio_take({:route, route}), do: @route_take_lookup[route]
   def audio_take(atom) when is_atom(atom), do: @atom_take_lookup[atom]
+
+  @spec paginate_text(String.t(), integer()) :: [{String.t(), String.t(), integer()}]
+  def paginate_text(text, max_length \\ 24) do
+    String.split(text)
+    |> Stream.chunk_while(
+      nil,
+      fn word, acc ->
+        if is_nil(acc) do
+          {:cont, word}
+        else
+          new_acc = acc <> " " <> word
+
+          if String.length(new_acc) > max_length do
+            {:cont, acc, word}
+          else
+            {:cont, new_acc}
+          end
+        end
+      end,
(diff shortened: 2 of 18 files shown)


lib/pa_ess/utilities.ex
```diff
@@ -697,4 +697,28 @@ defmodule PaEss.Utilities do
...
+            {:cont, acc, word}
+          else
+            {:cont, new_acc}
+          end
+        end
+      end,
+      fn acc -> {:cont, acc, nil} end
+    )
+    |> Stream.chunk_every(2, 2, [""])
+    |> Enum.map(fn [top, bottom] -> {top, bottom, 3} end)
```
sloanelybutsurely, line 722: **question:** what is the `, 3}` for here?

test/pa_ess/utilities_test.exs
```diff
@@ -109,4 +109,11 @@ defmodule Content.Audio.UtilitiesTest do
       assert replace_abbreviations("SvC, OK") == "Service, OK"
     end
   end
+
+  test "paginate_text" do
+    assert [{"Attention passengers:", "the next Braintree train", 3}, {"is now arriving", "", 3}] =
+             paginate_text("Attention passengers: the next Braintree train is now arriving", 24)
+
+    assert [{"too-long", "word", 3}] = paginate_text(" too-long   word ", 5)
+  end
```
sloanelybutsurely, line 118: **suggestion:** can we add a test where the message fits within the `max_length`?

sloanelybutsurely (commented):
**question (non-blocking):** do we have tests for things in `lib/content`? should we?

lib/content/audio.ex
```diff
@@ -19,6 +19,6 @@ defprotocol Content.Audio do
   @doc "Converts an audio struct to the mid/vars params for the PA system"
   @spec to_params(Content.Audio.t()) :: value()
   def to_params(audio)
-  @spec to_tts(Content.Audio.t()) :: String.t()
+  @spec to_tts(Content.Audio.t()) :: {String.t(), [{String.t(), String.t(), integer()}] | nil}
```
cmaddox5, line 22: question: Is the idea to phase this function out? Looks like all calls to it were replaced with `tts_text` calls, so mostly just making sure that was intentional.

lib/pa_ess/utilities.ex
panentheos: That will be the paging interval (when this actually gets used).

lib/content/audio.ex
panentheos: I should have mentioned in the description: This function isn't currently used, but it will be in a subsequent PR that implements the code path for sending to Scully.

panentheos:
Re: content tests. There are some, but they've proven to be pretty low-value since they're testing against a low-level, internal interface (which has also been shifting a lot recently). There's an extensive suite of high level behavior tests which exercise the sign algorithm end-to-end, including the low-level content code, and those have been much more useful for detecting actual changes in behavior. See `realtime_test.exs`. The plan is to expand those tests to assert on the new TTS outputs as well, once things are wired up a bit more.

sloanelybutsurely (approved):

PR #756, opened 2024-05-29
don't display predictions with skipped schedule_relationship

panentheos:
#### Summary of changes
**Asana Ticket:** [Update RTS filtering for non-revenue trains](https://app.asana.com/0/1185117109217413/1207429871853193/f)

This filters out predictions with `schedule_relationship` set to `SKIPPED`, to avoid displaying non-revenue predictions now that they include a departure time.


PaulJKim (approved):
Looks good!

PR #759, opened 2024-06-04
Modify "End of service" text to specify routes and/or directions

sloanelybutsurely:
#### Summary of changes
**Asana Ticket:** [Modify "End of service" text to specify routes and/or directions](https://app.asana.com/0/1185117109217422/1207279074432055/f)

Updates Last Trip messages to add additional context w/r/t route or direction when possible.

#### Reviewer Checklist
- [x] ~~Any new or changed functions have typespecs~~ No changes to functions

### DIFF
--- a/lib/content/message/last_trip/platform_closed.ex
+++ b/lib/content/message/last_trip/platform_closed.ex
@@ -11,7 +11,7 @@ defmodule Content.Message.LastTrip.PlatformClosed do
 
   defimpl Content.Message do
     def to_string(%Content.Message.LastTrip.PlatformClosed{}) do
-      "Platform closed"
+      "Service ended"
     end
   end
 end
--- a/lib/signs/utilities/last_trip.ex
+++ b/lib/signs/utilities/last_trip.ex
@@ -67,7 +67,7 @@ defmodule Signs.Utilities.LastTrip do
              not (is_prediction?(top_message) or is_prediction?(bottom_message)),
            do:
              {%LastTrip.PlatformClosed{destination: source.headway_destination},
-              %LastTrip.ServiceEnded{}},
+              %LastTrip.ServiceEnded{destination: source.headway_destination}},
            else: messages
     end
   end
--- a/lib/signs/utilities/last_trip.ex
+++ b/lib/signs/utilities/last_trip.ex
@@ -12,11 +12,7 @@ defmodule Signs.Utilities.LastTrip do
         {unpacked_mz_top, unpacked_mz_bottom} = unpack_mezzanine_content(messages, source)
         {top_source_config, bottom_source_config} = source
 
-        routes =
-          Enum.flat_map(
-            top_source_config.sources ++ bottom_source_config.sources,
-            &List.wrap(&1.routes)
-          )
+        routes = Signs.Utilities.SourceConfig.sign_routes(source)
 
         cond do
           # If combined alert status, only switch to Last Trip messaging once service has fully ended.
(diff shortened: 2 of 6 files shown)


lib/content/message/last_trip/platform_closed.ex
```diff
@@ -11,7 +11,7 @@ defmodule Content.Message.LastTrip.PlatformClosed do
 
   defimpl Content.Message do
     def to_string(%Content.Message.LastTrip.PlatformClosed{}) do
-      "Platform closed"
+      "Service ended"
```
sloanelybutsurely, line 14: **question:** should this be a different struct altogether? should i combine it with `%ServiceEnded{}` (which displays "Service ended for night" or "No <headsign> trains") with some sort of additional configuration?

if we use these as representations of what _kind_ of content to show, sort of the context of the message, i feel like we can keep this as-is?

lib/signs/utilities/last_trip.ex
```diff
@@ -10,7 +10,13 @@ defmodule Signs.Utilities.LastTrip do
...
       {has_top_service_ended?, has_bottom_service_ended?} ->
         {unpacked_mz_top, unpacked_mz_bottom} = unpack_mezzanine_content(messages, source)
-        {top_source, bottom_source} = source
+        {top_source_config, bottom_source_config} = source
+
+        routes =
+          Enum.flat_map(
+            top_source_config.sources ++ bottom_source_config.sources,
+            &List.wrap(&1.routes)
+          )
```
PaulJKim, line 19: We have a helper method [here](https://github.com/mbta/realtime_signs/blob/main/lib/signs/utilities/source_config.ex#L193-L195) that should let you do something like
```suggestion
        routes = Signs.Utilities.SourceConfig.sign_routes(source)
```

lib/content/message/last_trip/platform_closed.ex
PaulJKim: Yeah, I agree we don't need to combine it with the `%ServiceEnded{}` struct. Like you said, the struct names just express the context of the message, which isn't changing, and whatever text they actually render to is just a minor detail imo.

test/signs/realtime_test.exs
```diff
@@ -1406,9 +1424,20 @@ defmodule Signs.RealtimeTest do
...
     end
 
+    test "Station with routes is closed" do
+      sign = %{
+        @mezzanine_sign_with_routes
+        | tick_read: 0
+      }
+
+      expect_messages({"No Orange Line", "Service ended for night"})
+      expect_audios(canned: {"103", ["883"], :audio})
```
PaulJKim, line 1434: `883` reads out "This station is closed. Service has ended for the night.", but the ticket says it should now read "Orange Line service has ended for the night". Luckily, we already have take `882` that reads just "service has ended for the night" which can combine with takes that we have for the route names. We'll have to start passing the routes to `%Content.Audio.ServiceEnded{}` though and call a helper within its `to_params/1` implementation for the bidirectional case.

sloanelybutsurely: Oh nice catch! How does this look? https://github.com/mbta/realtime_signs/pull/759/commits/a799010fd5209f50bdd875b8594fec321d9f6851

PaulJKim: Overall seems right! But I think `routes` will always be a list though so could we get away with removing the call to `List.wrap/1`? 

I'm also a little torn on whether or not we should default to using `@service_ended`. The `Utilities.line_to_var()` helper defaults to using the take for "Train" which I think could be fine.

sloanelybutsurely: @PaulJKim okay, updated to remove `List.wrap` and defaulting to `@service_ended`: [`a799010..137cc1d`](https://github.com/mbta/realtime_signs/compare/a799010fd5209f50bdd875b8594fec321d9f6851..137cc1d427477a8d484281bc1d150d259030d33a?w=1)

PaulJKim (approved):
Looks good!

PR #764, opened 2024-06-11
Trigger PA Messages from RTS

PaulJKim:
#### Summary of changes
**Asana Ticket:** [RTS: Trigger PA Message Readouts](https://app.asana.com/0/1185117109217413/1207286579042501/f)

This PR adds a new engine that reads in active PA messages from Screenplay and schedules them to play on sign zones. It also adds a map to sign states to track when the most recent play of a given pa message was. This is to add some resilience against engine restarts which would result in triggers of PA messages that are too close together.

-When a message goes inactive, it's current timer will be cancelled and it's record delete from the ets table.
-When an interval is changed, we will update the scheduling so that the next time the message plays will be in line with it's new interval.


lib/signs/realtime.ex
```diff
@@ -47,7 +47,8 @@ defmodule Signs.Realtime do
                 announced_alert: false,
                 prev_prediction_keys: nil,
                 prev_predictions: [],
-                uses_shuttles: true
+                uses_shuttles: true,
+                pa_message_plays: %{}
```
panentheos, line 51: I'm not sure I understand the benefit of tracking this here. Is this so in case the engine crashes, we won't spam messages? If that's the concern, and we're willing to accept a slight behavior change, we could schedule new messages with their full duration instead of immediately, which would result in fewer messages in that failure mode, rather than excess ones.

config/runtime.exs
```diff
@@ -36,7 +36,10 @@ if config_env() != :test do
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
     message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
     monitoring_api_key: System.get_env("MONITORING_API_KEY"),
-    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH")
+    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH"),
+    screenplay_url: System.get_env("SCREENPLAY_URL"),
+    screenplay_api_key: System.get_env("SCREENPLAY_API_KEY"),
+    active_pa_messages_path: System.get_env("ACTIVE_PA_MESSAGES_PATH")
```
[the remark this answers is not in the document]
PaulJKim: Hmm yeah that's a good point. I think my rationale was just that the path could change since the screenplay api is relatively new and still being developed, but I guess it would be just as much effort to make the devops change rather than just make a code change in RTS or probably even less. I'll make a change to remove this one

lib/engine/pa_messages.ex
```diff
@@ -0,0 +1,191 @@
...
+        when existing_pa_message.interval_in_ms != active_pa_message.interval_in_ms ->
+          case Process.read_timer(timer_ref) do
+            false ->
+              schedule_pa_message(active_pa_message, active_pa_message.interval_in_ms, table)
+
+            remaining_ms ->
+              ms_elapsed = existing_pa_message.interval_in_ms - remaining_ms
+              temp_interval = (active_pa_message.interval_in_ms - ms_elapsed) |> max(0)
+              cancel_pa_timer(timer_ref, pa_id)
+              schedule_pa_message(active_pa_message, temp_interval, table)
```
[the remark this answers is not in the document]
PaulJKim: Hmm yeah I think that approach works nicely and will make the "edited interval" piece of the logic a good deal similar. And it removes the need to cancel timers and such as you hinted at. I think we also would get the "handling" of inactive messages for free as well since if they're not included in the response, we won't be checking for the last playtime anyway.

lib/signs/realtime.ex
PaulJKim: Yeah that was the intention here, just to prevent consecutive plays due to an engine crashes/restarts. Ah so you mean that a PA Message wouldn't play immediately upon scheduling and instead wait one interval before it's first play?

I had actually sort of [brought that up](https://mbta.slack.com/archives/C039ARUM48H/p1717691404196349) in the product channel last week but it was only discussed briefly. Kevin said a PA message should play as soon as it can right after it's created, but I'm still not sure if that's also what the ARINC software does currently. Seems like its still certainly up for discussion though. I'll raise the question again to the product folks and see if we would be okay with that behavior change.

PaulJKim: Discussed a bit further in the thread and I think I lean towards avoiding scheduling a new message with the full duration. I don't think it's been fully confirmed or is explicitly a standard practice, but it seems that OIOs tend to schedule a message and look immediately for a "proof of play" and I'd rather not disrupt their workflow if we can avoid it. I'm open to other ideas for preventing spammed message plays though if you have any thoughts?

config/runtime.exs
```diff
@@ -36,7 +36,9 @@ if config_env() != :test do
     message_log_s3_folder: System.get_env("MESSAGE_LOG_S3_FOLDER"),
     message_log_report_s3_folder: System.get_env("MESSAGE_LOG_REPORT_S3_FOLDER"),
     monitoring_api_key: System.get_env("MONITORING_API_KEY"),
-    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH")
+    active_headend_path: System.get_env("ACTIVE_HEADEND_S3_PATH"),
+    active_pa_messages_url: System.get_env("ACTIVE_PA_MESSAGES_URL"),
```
PaulJKim, line 40: On second thought @panentheos , I'm not sure if I interpreted your comment about the runtime env here correctly. Did you mean that it's not necessary to split the screenplay base url and the route for active PA messages or that we don't need the screenplay url as an env variable at all? I assume we still want it as an env variable for local testing and such, right?

panentheos: My thinking was: The screenplay base URL should be an env variable, like we do for other systems. The path for the endpoint should not be: Ideally, it wouldn't be abstracted at all, since the actual API path on Screenplay is always the same. However, because of how the `Fake` module works, we need some sort of hook for testing different responses. The least intrusive thing that gets us that would be to put the path in the _application_ environment, with the real value configured statically in `config.exs`, and then have the tests modify the application env like they're doing.

lib/signs/utilities/audio.ex
```diff
@@ -402,4 +402,25 @@ defmodule Signs.Utilities.Audio do
...
+
+      last_sent ->
+        if DateTime.diff(DateTime.utc_now(), last_sent, :millisecond) >= pa_message.interval_in_ms do
+          send_audio_fn.()
+        else
+          Logger.warn("pa_message: action=skipped id=#{pa_message.id}")
+        end
+    end
+    |> then(fn
+      :ok -> Map.put(pa_message_plays, pa_message.id, DateTime.utc_now())
```
panentheos, line 422: `PaEss.Updater.play_message` doesn't currently return `:ok` for the new SCU code path, so it should be changed to do that. Alternatively, we could just remove the branching here, since I don't think we expect a recoverable failure case anyway.

lib/engine/pa_messages.ex
```diff
@@ -0,0 +1,141 @@
...
+      Logger.info("pa_message: action=send id=#{pa_message.id} destination=#{sign_id}")
+
+      send(
+        String.to_existing_atom("Signs/#{sign_id}"),
+        {:play_pa_message, pa_message}
+      )
+    end)
+  end
+
+  defp handle_inactive_pa_messages(active_pa_messages, pa_messages_last_sent) do
```
panentheos, line 98: Do we want to do this? If we proactively wipe this state, then deactivating and reactivating an alert would cause it to play again immediately, which could be undesirable. If we leave the sign state intact, it would block it instead.

lib/signs/bus.ex
```diff
@@ -121,6 +124,23 @@ defmodule Signs.Bus do
     {:ok, state}
   end
 
+  @impl true
+  def handle_info({:play_pa_message, pa_message}, sign) do
+    Logger.info("pa_message: action=send id=#{pa_message.id} destination=#{sign.id}")
```
panentheos, line 129: The logging looks like it could be cleaned up a little. This one should probably be moved into `send_pa_message` and only done if we actually play the message. There's another log with the same `action` value in the engine, which we could remove.

panentheos (approved):
See a few comments, then looks good. Before we merge/deploy this, we should make sure that the environment variables are set up, and that screenplay is ready to serve responses in prod.

lib/engine/pa_messages.ex
PaulJKim: Yeah that makes sense. What I had in mind was to keep things in line with the immediate feedback that OIOs tend to expect when they set up a PA message, but maybe that's just for initially creating the message or reactivating it after a longer length of time has passed. Probably better to not allow the chance of a bunch of consecutive plays.

PaulJKim: I could see the logging here being a useful audit, but we could also just do that from the Screenplay side

PR #765, opened 2024-06-11
chore: install and configure remote_ip

sloanelybutsurely:
See also https://github.com/mbta/screenplay/pull/366


digitalcora (approved):
📡

PR #766, opened 2024-06-11
add routes to test signs

panentheos:
#### Summary of changes

This adds route data to the sign structures in the tests, which makes them more reflective of the actual sign configs, and eliminates the need for a separate case for this.

### DIFF
--- a/lib/signs/utilities/source_config.ex
+++ b/lib/signs/utilities/source_config.ex
@@ -96,13 +96,14 @@ defmodule Signs.Utilities.SourceConfig do
   @enforce_keys [
     :stop_id,
     :direction_id,
+    :routes,
     :platform,
     :terminal?,
     :announce_arriving?,
     :announce_boarding?
   ]
   defstruct @enforce_keys ++
-              [:routes, :headway_stop_id, multi_berth?: false]
+              [:headway_stop_id, multi_berth?: false]
 
   @type source :: %__MODULE__{
           stop_id: String.t(),
--- a/test/signs/realtime_test.exs
+++ b/test/signs/realtime_test.exs
@@ -8,6 +8,7 @@ defmodule Signs.RealtimeTest do
   @src %Signs.Utilities.SourceConfig{
     stop_id: "1",
     direction_id: 0,
+    routes: ["Red"],
     platform: nil,
     terminal?: false,
     announce_arriving?: true,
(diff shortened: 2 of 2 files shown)


test/signs/realtime_test.exs
```diff
@@ -8,6 +8,7 @@ defmodule Signs.RealtimeTest do
   @src %Signs.Utilities.SourceConfig{
     stop_id: "1",
     direction_id: 0,
+    routes: ["Red"],
```
digitalcora, line 11: Is there a case where this is ever absent? Should we actively prevent it from being absent, using `@enforce_keys`?

panentheos: Good call, done.

lib/signs/utilities/source_config.ex
```diff
@@ -96,13 +96,14 @@ defmodule Signs.Utilities.SourceConfig do
   @enforce_keys [
     :stop_id,
     :direction_id,
+    :routes,
```
digitalcora, line 99: The typespec also says this could be `nil`, which maybe now we don't expect it to be? (unless there are places we specifically set it to `nil`)

test/signs/realtime_test.exs
```diff
@@ -244,33 +228,33 @@ defmodule Signs.RealtimeTest do
 
     test "when sign is at a station closed by shuttles and there are no predictions, it says so" do
       expect(Engine.Alerts.Mock, :max_stop_status, fn _, _ -> :shuttles_closed_station end)
-      expect_messages({"No train service", "Use shuttle bus"})
```
digitalcora, line 247: Does there remain any way to test the "route-less" logic now? Is it dead code?

digitalcora (approved):
Assuming a response to current comments one way or the other, looks good!

lib/signs/utilities/source_config.ex
panentheos: Below this gets populated from the JSON in a way that makes `nil` possible. I don't think we actually (want to) do that, but until we can verify that assumption, probably best to leave it like that.

test/signs/realtime_test.exs
panentheos: Almost, but not quite. The Ashmont mezzanine can still trigger this case. I added another case to exercise that.

digitalcora: Nice, even better test coverage then!

PR #767, opened 2024-06-11
fix several audio issues

panentheos:
#### Summary of changes

This fixes three audio issues:
1. When partial early AM suppression on a mezzanine sign results in two headway messages, the resulting combined message ends up with both a `destination` and `routes`. The headway TTS audio code wasn't accounting for that possibility, resulting in a crash.
2. In the same case as above, the non-TTS audio code was using a slightly different algorithm than the text code, resulting in a headway readout using the top line's destination, instead of the route name.
3. The TTS audio code for station-level service ended messages was still saying "This station is closed" instead of using the line name.

### DIFF
--- a/lib/content/audio/service_ended.ex
+++ b/lib/content/audio/service_ended.ex
@@ -64,8 +64,9 @@ defmodule Content.Audio.ServiceEnded do
       {tts_text(audio), nil}
     end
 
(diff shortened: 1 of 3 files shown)


cmaddox5 (approved):
Looks good to me.

PR #780, opened 2024-06-20
feat: SL Headway Mode

cmaddox5:
#### Summary of changes
**Asana Ticket:** [Add headway mode for Silver Line signs](https://app.asana.com/0/1185117109217413/1207396498133396/f)

This PR prevents SL signs from switching to Auto when Headway mode is on. Headway mode on SL signs is purely for suppressing predictions. There is no visual component to it.

To be merged with: https://github.com/mbta/signs_ui/pull/1344

#### Reviewer Checklist
- [X] Any new or changed functions have typespecs
- [X] Tests were added for any new functionality (don't just rely on Codecov)
- [X] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes (compare on Splunk: [staging](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-dev%20%22%5Berror%5D%22%20OR%20%22%5Bwarn%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840107.3874236) vs. [prod](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-prod%20%22%5Berror%5D%22%20OR%20%22%5Bwarn%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840137.3874305))


lib/signs/bus.ex
```diff
@@ -841,4 +841,9 @@ defmodule Signs.Bus do
       )
     end
   end
+
+  # If a Silver Line sign is in headway mode, SignsUI will flag its predictions but RTS needs to treat it as off
+  defp config_off?("Silver_Line." <> _, :headway), do: true
```
digitalcora, line 846: I don't have much of the context on RTS yet, but is there no more-robust way to tell whether a sign is part of the Silver Line group? (if "group" is even the right word) I'm wary of peeking into the contents of IDs.

digitalcora (commented):
There was some discussion in the task for this, which said:

> Currently Transit Data produces predictions for the Silver Line Transitway stations, but we do not show them.

I asked if we should also remove the code that handles this "we do not show them" status quo, and Dave said yes, it should be safe to remove already as the relevant changes are already live on the Transit Data side. So assuming that suppression does exist somewhere in RTS, can we remove it?

lib/signs/bus.ex
cmaddox5: Unfortunately we don't have much: https://github.com/mbta/signs_ui/blob/f3c647b800a380649014f67eea97b1f9d6ff4abe/priv/signs.json#L8554. SL and bus share the same config shape so it's tough to tell them apart aside from the IDs. Maybe the best thing to do here is peak into `sources` and make sure that all `route_id`s are SL route IDs? I'm willing to bet an SL route ID change is a lot less likely than an ID overhaul so the code should be more stable.

cmaddox5:
> There was some discussion in the task for this, which said:
> 
> > Currently Transit Data produces predictions for the Silver Line Transitway stations, but we do not show them.
> 
> I asked if we should also remove the code that handles this "we do not show them" status quo, and Dave said yes, it should be safe to remove already as the relevant changes are already live on the Transit Data side. So assuming that suppression does exist somewhere in RTS, can we remove it?

Sorry for missing this. Will dig in a bit more to find where this might be happening. The only bit of "don't show them" logic I found so far is in [bus.ex](https://github.com/mbta/realtime_signs/blob/e2bf8357d1409fd04f976a093eb2fd7a201c97a1/lib/signs/bus.ex#L278) where we filter out predictions that are too far in the future. Without the `Stream.reject` piece, the predictions show like they normally would. Will dig in a bit more to make sure I'm not missing anything.

lib/signs/bus.ex
cmaddox5: I think I found a way to make that work. We were already fetching all of the route IDs for alert purposes, so I just reused that value.

test/signs/bus_test.exs
```diff
@@ -71,6 +71,7 @@ defmodule Signs.BusTest do
   defmodule FakeConfig do
     def sign_config("auto_sign"), do: :auto
     def sign_config("off_sign"), do: :off
+    def sign_config("Silver_Line.headway"), do: :headway
```
digitalcora, line 74: This doesn't need this specific ID format anymore, right?

lib/signs/bus.ex
```diff
@@ -9,6 +9,7 @@ defmodule Signs.Bus do
   @drawbridge_minutes_spanish "152"
   @drawbridge_soon_spanish "157"
   @alert_levels [:station_closure, :suspension_closed_station]
+  @sl_route_ids MapSet.new(["741", "742", "743", "746"])
```
digitalcora, line 12: Should we include `749` and `751` here for completeness? Or maybe call this `@sl_waterfront_route_ids`.

digitalcora (approved):
Pre-approving with minor comments, looks good as far as I can see.

test/signs/bus_test.exs
cmaddox5: Correct. I'll change it to just be `headway`.

lib/signs/bus.ex
cmaddox5: I think a more specific name is better here. `749` and `751` only show on Nubian and those are not configurable on the SL page in SignsUI.

PR #781, opened 2024-06-20
Clear Last Trip related data at 3:30am

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Modify logic for LTOTD](https://app.asana.com/0/1185117109217413/1207279074432059/f)

Instead of expiring last trip ids and recent departures for 2 hours and 90 minutes respectively, we decided that it would be better to just lets Last Trip messaging stay up and then turn off all at once at 3:30am each day.


sloanelybutsurely (approved):
lgtm

PR #785, opened 2024-06-25
add bus TTS audio

panentheos:
#### Summary of changes
**Asana Ticket:** [Support communication with new SCU app](https://app.asana.com/0/1185117109217413/1206141631184125/f)

This adds TTS versions of bus audio messages. These mirror the current audio message formats, except for the Spanish bridge message, which currently incorrectly shows English text.

### DIFF
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -678,7 +678,7 @@ defmodule Signs.Bus do
       {duration, duration_spanish} =
         case bridge_status_minutes(bridge_status, current_time) do
           minutes when minutes < 2 ->
-            {"We expect it to be lowered soon.", "Esperamos que baje pronto."}
+            {"We expect it to be lowered soon.", "Esperamos que se baje pronto."}
 
           minutes ->
             {"We expect this to last for at least #{minutes} more minutes.",
@@ -689,7 +689,7 @@ defmodule Signs.Bus do
         "The Chelsea Street bridge is raised. #{duration} SL3 buses may be delayed, detoured, or turned back."
 
       spanish_text =
-        "Se levanta el puente de Chelsea Street. #{duration_spanish} Los autobuses SL3 pueden sufrir retrasos, desvíos o dar marcha atrás."
+        "El puente de Chelsea Street está levantado. #{duration_spanish} Los autobuses SL3 pueden sufrir retrasos, desvíos o dar marcha atrás."
 
       [
         {english_text, PaEss.Utilities.paginate_text(english_text)},
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -121,6 +121,10 @@ defmodule Signs.Bus do
     {:ok, state}
   end
 
+  @type content_values ::
+          {messages :: [Content.Message.value()], audios :: [Content.Audio.value()],
+           tts_audios :: [Content.Audio.tts_value()]}
+
   @impl true
   def handle_info(:run_loop, state) do
     Process.send_after(self(), :run_loop, 1000)
@@ -168,7 +172,10 @@ defmodule Signs.Bus do
     {[top, bottom], audios, tts_audios} =
       cond do
         config == :off ->
-          {["", ""], [], []}
+          messages = ["", ""]
+          audios = []
+          tts_audios = []
+          {messages, audios, tts_audios}
 
         match?({:static_text, _}, config) ->
           static_text_content(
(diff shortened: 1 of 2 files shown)


lib/signs/bus.ex
```diff
@@ -165,10 +165,10 @@ defmodule Signs.Bus do
       |> Enum.group_by(&{&1.stop_id, &1.route_id, &1.direction_id})
 
     # Compute new sign text and audio
-    {[top, bottom], audios} =
+    {[top, bottom], audios, tts_audios} =
       cond do
         config == :off ->
-          {["", ""], []}
+          {["", ""], [], []}
```
digitalcora, line 171: This complex data type formed of anonymous lists and tuples is becoming a bit unwieldy, I think. Is it time to introduce a struct with named fields (and, ideally, a typespec)?

panentheos: I think a struct would be too heavy here, since these values are just being returned to the outer scope and then immediately used. I added some types to document things a bit, and also added some explicit variables to this line (which otherwise looks particularly odd because of the empty values)

digitalcora: Nice, the type definitely helps.

We could also keep the variable approach concise like this, if you want:

```elixir
{_messages = ["", ""], _audios = [], _tts_audios = []}
```

lib/signs/bus.ex
```diff
@@ -612,6 +651,36 @@ defmodule Signs.Bus do
...
 
+  defp bridge_tts_audio(bridge_status, bridge_enabled?, current_time, state) do
+    %{chelsea_bridge: chelsea_bridge} = state
+
+    if bridge_enabled? && chelsea_bridge &&
+         bridge_status_raised?(bridge_status, current_time) do
+      {duration, duration_spanish} =
+        case bridge_status_minutes(bridge_status, current_time) do
+          minutes when minutes < 2 ->
+            {"We expect it to be lowered soon.", "Esperamos que se reduzca pronto."}
```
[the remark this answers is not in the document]
panentheos: Good catch. I went with `baje` as opposed to `se baje`, according to Translate. I think `se` was acting as "to be" in the previous one, but `baje` is an active verb so doesn't need it.

lib/signs/bus.ex
panentheos: Oh interesting, I didn't realize that was possible. Thanks.

lib/signs/bus.ex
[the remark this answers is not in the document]
digitalcora: `se` is actually the pronoun here, so I'm pretty sure it should remain in place (you can translate this literally as "<ruby>We expect<rt>Esperamos</rt></ruby> / <ruby>that<rt>que</rt></ruby> / <ruby>it<rt>se</rt></ruby> / <ruby>will be lowered<rt>baje</rt></ruby> / <ruby>soon<rt>pronto</rt></ruby>"). "se + verb" is the passive voice — no specified entity will lower the bridge, it just "will be lowered".

I'm admittedly not sure why `baje` is the right conjugation to use here (it's the "subjunctive" and the rules around that are not simple), but our respective machine translations seem to agree it's the right one.

lib/signs/bus.ex
```diff
@@ -612,6 +670,36 @@ defmodule Signs.Bus do
...
+          minutes ->
+            {"We expect this to last for at least #{minutes} more minutes.",
+             "Esperamos que esto dure al menos #{minutes} minutos más."}
+        end
+
+      english_text =
+        "The Chelsea Street bridge is raised. #{duration} SL3 buses may be delayed, detoured, or turned back."
+
+      spanish_text =
+        "Se levanta el puente de Chelsea Street. #{duration_spanish} Los autobuses SL3 pueden sufrir retrasos, desvíos o dar marcha atrás."
```
digitalcora, line 692: On the topic of conjugations, this should probably be `se levante` — I think "se levant**a** el puente" means something like "the bridge raises" (is currently going up).

I plugged this into DeepL as well and it suggested _"El puente de Chelsea Street está levantado"_, which might be a more natural construction. `levantado` is the past participle (adjective) form of the verb, making this a fairly direct translation: <ruby>The bridge<rt>El puente</rt></ruby> / <ruby>of<rt>de</rt></ruby> / Chelsea Street / <ruby>is<rt>está</rt></ruby> / <ruby>raised<rt>levantado</rt></ruby>.

panentheos:
We should probably run this by a Spanish speaker at some point before go-live.

digitalcora:
Yes, I think we do have some people around who'd be able to do better than my extremely rusty high-school Spanish plus machine translation 😄

digitalcora (approved):
Looks good to me! If we handle final Spanish translation as a separate task, let's just make sure that's tracked somewhere.

PR #793, opened 2024-07-01
remove phoenix_view and phoenix_html

panentheos:
#### Summary of changes

RTS has no visual web endpoints, so we're not using these libraries at all. We can remove them instead of upgrading them.


sloanelybutsurely (approved):
␡

mix.exs
```diff
@@ -56,9 +56,7 @@ defmodule RealtimeSigns.Mixfile do
       {:uuid, "~> 1.1", only: :test},
       {:quantum, "~> 3.0"},
       {:phoenix, "~> 1.7.0"},
-      {:phoenix_html, "~> 3.0"},
       {:phoenix_live_reload, "~> 1.2", only: :dev},
```
digitalcora, line 59: Don't need this one anymore either, I think?

panentheos: There's still a reference to it, but I guess we could strip that out too.

PR #795, opened 2024-07-09
RTS update_signs: logs include message_type for custom or headway modes

sloanelybutsurely:
#### Summary of changes
**Asana Ticket:** [RTS update_signs: logs include message_type for custom or headway modes](https://app.asana.com/0/1185117109217413/1204643930705430/f)

- fetches sign config and includes message_type in `update_sign` logging

#### Reviewer Checklist

### DIFF
--- a/lib/pa_ess/http_updater.ex
+++ b/lib/pa_ess/http_updater.ex
@@ -83,7 +83,13 @@ defmodule PaEss.HttpUpdater do
       signs_ui_ms: signs_ui_ms,
       top_line: inspect(top_line),
       bottom_line: inspect(bottom_line),
-      sign_id: sign_id
+      sign_id: sign_id,
+      message_type:
+        case Engine.Config.sign_config(sign_id) do
+          type when is_atom(type) -> type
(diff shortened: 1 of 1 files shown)


lib/pa_ess/http_updater.ex
```diff
@@ -83,7 +83,13 @@ defmodule PaEss.HttpUpdater do
       signs_ui_ms: signs_ui_ms,
       top_line: inspect(top_line),
       bottom_line: inspect(bottom_line),
-      sign_id: sign_id
+      sign_id: sign_id,
+      message_type:
+        case Engine.Config.sign_config(sign_id) do
```
panentheos, line 88: This introduces a slight chance of a race condition, if a sign's config changes between when the message is queued and here. That might not be a big deal since this is just used for debugging, but it's worth mentioning. If we want to avoid that, we could do something similar to the audio code path, and add an `extra_logs` value alongside the rest of the queued data, which is fed from the actual data used to generate the message (but that would be slightly more complicated).

If we do decide to go this route, I think something like `current_config` would be a better name than `message_type` for the log key.

PR #796, opened 2024-07-15
chore: add lab sign config

digitalcora:
### [chore: add lab sign config](https://github.com/mbta/realtime_signs/pull/796/commits/4c86c2ebb11880d0fffd97b23d7268a342bff398)

This adds a definition for a "test" sign(s) set up in our office. Other than the unique IDs, the config is a copy of `state_blue_mezzanine` for now.

### [feat: signs can have a "default mode"](https://github.com/mbta/realtime_signs/pull/796/commits/9f007971d60b331bbb0cd90a2386fab90e500e73)

Previously, if a sign wasn't configured in SignsUI, it would always be in the `off` mode. We'd like to be able to configure our new lab signs for testing without the complexity of supporting this unique "kind" of sign in SignsUI as well as this project.

This adds an optional `default_mode` field to sign configuration, which can be set to any of the "modes" supported by SignsUI. For now, this is only supported by `Realtime` (a.k.a. subway) signs and not `Bus` signs, but this would be easy enough to extend later if needed.

Due to now needing a sign's default mode as well as its ID whenever we want to determine its "current mode", this includes a refactor of the previously-added `current_config` logging so the value is determined when a message is enqueued, instead of when it is later processed.

#### Reviewer Checklist
- [x] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes (compare on Splunk: [staging](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-dev%20%22%5Berror%5D%22%20OR%20%22%5Bwarn%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840107.3874236) vs. [prod](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-prod%20%22%5Berror%5D%22%20OR%20%22%5Bwarn%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840137.3874305))

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207677723099981

### DIFF
--- a/lib/engine/config.ex
+++ b/lib/engine/config.ex
@@ -32,10 +32,10 @@ defmodule Engine.Config do
   end
 
   @impl true
-  def sign_config(table_name \\ @table_signs, sign_id) do
-    case :ets.lookup(table_name, sign_id) do
-      [{^sign_id, config}] -> config
-      _ -> :off
+  def sign_config(table \\ @table_signs, sign_id, default) do
+    case :ets.lookup(table, sign_id) do
+      [{^sign_id, config}] when not is_nil(config) -> config
+      _ -> default
     end
   end
 
@@ -200,7 +200,7 @@ defmodule Engine.Config do
         :temporary_terminal
 
       true ->
-        :off
+        nil
     end
   end
 
--- a/lib/signs/realtime.ex
+++ b/lib/signs/realtime.ex
@@ -45,6 +45,7 @@ defmodule Signs.Realtime do
                 announced_stalls: [],
                 announced_custom_text: nil,
                 announced_alert: false,
+                default_mode: :off,
                 prev_prediction_keys: nil,
                 prev_predictions: [],
                 uses_shuttles: true,
@@ -64,6 +65,7 @@ defmodule Signs.Realtime do
           text_zone: String.t(),
           audio_zones: [String.t()],
           source_config: SourceConfig.config() | {SourceConfig.config(), SourceConfig.config()},
+          default_mode: Engine.Config.sign_config(),
           current_content_top: Content.Message.value(),
           current_content_bottom: Content.Message.value(),
           prediction_engine: module(),
@@ -101,6 +103,8 @@ defmodule Signs.Realtime do
       text_zone: Map.fetch!(config, "text_zone"),
       audio_zones: Map.fetch!(config, "audio_zones"),
       source_config: source_config,
+      default_mode:
+        config |> Map.get("default_mode") |> then(&if(&1 == "auto", do: :auto, else: :off)),
       current_content_top: "",
       current_content_bottom: "",
       prediction_engine: Engine.Predictions,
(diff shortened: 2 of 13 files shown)


priv/signs.json
```diff
@@ -1,4 +1,53 @@
...
+    "id": "lab_test",
+    "type": "realtime",
+    "pa_ess_loc": "201",
+    "scu_id": "TBR01",
+    "read_loop_offset": 120,
+    "text_zone": "m",
+    "audio_zones": [
+      "m",
+      "n",
+      "s"
```
digitalcora, line 12: I'm not sure if including _all_ of the supported zones is what we want here. Should it be only a subset?

digitalcora:
This has been running on dev overnight and does not seem to have encountered any issues, though all it's doing is sending an empty string for both lines — I assume this is because the "sign mode" that for a normal sign would be configured in `signs_ui` defaults to "off". This task didn't mention anything about setting up `signs_ui` to handle the lab sign, but I'm not sure if that's an intentional omission.

priv/signs.json
panentheos: Just one is more typical, so let's use "m".

lib/engine/config.ex
```diff
@@ -32,10 +32,14 @@ defmodule Engine.Config do
   end
 
   @impl true
-  def sign_config(table_name \\ @table_signs, sign_id) do
-    case :ets.lookup(table_name, sign_id) do
-      [{^sign_id, config}] -> config
-      _ -> :off
+  def sign_config(table \\ @table_signs, sign)
+  def sign_config(table, %{id: id, default_mode: default}), do: lookup(table, id, default)
+  def sign_config(table, %{id: id}), do: lookup(table, id, :off)
```
panentheos, line 37: nit: There isn't an immediate need to implement this for bus signs, but generally we've been looking to nudge the rail and bus code closer together over time. So instead of having this be optional, we could hardcode a value of `:off` in the bus state, and then they could use the same code path.

lib/signs/realtime.ex
```diff
@@ -101,6 +103,16 @@ defmodule Signs.Realtime do
       text_zone: Map.fetch!(config, "text_zone"),
       audio_zones: Map.fetch!(config, "audio_zones"),
       source_config: source_config,
+      default_mode:
+        config
+        |> Map.get("default_mode")
+        |> then(fn
```
panentheos, line 109: nit: Since the need for this is so narrow, I don't think we need to exhaustively support all modes. I think we could just match `auto`, and have a default of `off`, and call it a day.

panentheos (approved):
See a few small comments, then looks good.

lib/signs/realtime.ex
digitalcora: I thought it might come in handy, if we wanted to mimic the effect of OIOs setting a particular mode in Signs UI, to support them all (since we cannot do this via Signs UI for the lab sign). Maybe I'm not quite understanding what the lab sign will be used for, though. If you're confident we won't need anything other than `auto`, then we can drop the rest.

panentheos: The main purpose of the test signs is the physical hardware itself, so we just need some kind of messages going to it. Any testing of the config logic we can easily do in dev with an existing sign and signs-ui, so all we need is a quick way to force auto mode here.

PR #797, opened 2024-07-17
fix: expand handling of alert status on signs with top/bottom configs

sloanelybutsurely:
#### Summary of changes
**Asana Ticket:** [Ashmont MZ sign shows headway mode for Mattapan Trolley when it's shuttling](https://app.asana.com/0/1185117109217413/1207563461018208/f)

different alert status per top/bottom config are needed to support showing an alert status based message on a single source config. this is most common when top and bottom source configs are for different lines/routes.

previously the alert_status for a sign was reduced down to a single value which could cause weird behavior when an alert was only effecting one of the two source configs.

https://github.com/user-attachments/assets/1ab76a7d-fe2e-4884-b0df-a1feb070402e

#### Reviewer Checklist

### DIFF
--- a/lib/signs/realtime.ex
+++ b/lib/signs/realtime.ex
@@ -146,11 +146,26 @@ defmodule Signs.Realtime do
   end
 
   def handle_info(:run_loop, sign) do
-    sign_stop_ids = SourceConfig.sign_stop_ids(sign.source_config)
-    sign_routes = SourceConfig.sign_routes(sign.source_config)
-    alert_status = sign.alerts_engine.max_stop_status(sign_stop_ids, sign_routes)
     sign_config = sign.config_engine.sign_config(sign.id, sign.default_mode)
     current_time = sign.current_time_fn.()
+
+    alert_status =
+      case sign.source_config do
+        {top, bottom} ->
+          top_alert_status = get_alert_status_for_source_config(top, sign.alerts_engine)
+          bottom_alert_status = get_alert_status_for_source_config(bottom, sign.alerts_engine)
+
+          {top_alert_status, bottom_alert_status}
+
+        source_config ->
+          alert_status = get_alert_status_for_source_config(source_config, sign.alerts_engine)
+
+          alert_status
+      end
+      |> then(fn
+        {alert_status, alert_status} -> alert_status
+        alert_status -> alert_status
+      end)
 
     first_scheduled_departures =
       case sign.source_config do
@@ -164,8 +179,9 @@ defmodule Signs.Realtime do
           }
 
         source ->
-          {sign.headway_engine.get_first_scheduled_departure(sign_stop_ids),
-           source.headway_destination}
+          {sign.headway_engine.get_first_scheduled_departure(
+             SourceConfig.sign_stop_ids(sign.source_config)
+           ), source.headway_destination}
       end
 
     prev_predictions_lookup =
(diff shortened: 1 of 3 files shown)


lib/signs/realtime.ex
```diff
@@ -142,12 +142,27 @@ defmodule Signs.Realtime do
...
+        {top, bottom} ->
+          top_alert_status = get_alert_status_for_source_config(top, sign.alerts_engine)
+          bottom_alert_status = get_alert_status_for_source_config(bottom, sign.alerts_engine)
+
+          {top_alert_status, bottom_alert_status}
+
+        source_config ->
+          alert_status = get_alert_status_for_source_config(source_config, sign.alerts_engine)
+
+          alert_status
```
PaulJKim, line 159: Doesn't seem like the `alert_status` variable is needed for anything in this scope, unless it's just a preference?

PaulJKim (commented):
Do we need to account for the possible split alert statuses when calling `get_alert_message` [here](https://github.com/mbta/realtime_signs/pull/797/files#diff-bdda4cc40319158f7ed70a48350c960125e1cc22848b0c5f9dccde12c05205beR39)? I think it would be possible for a MZ sign to have differing alert statuses for either line but be set to headway mode.

sloanelybutsurely:
> Do we need to account for the possible split alert statuses when calling `get_alert_message` [here](https://github.com/mbta/realtime_signs/pull/797/files#diff-bdda4cc40319158f7ed70a48350c960125e1cc22848b0c5f9dccde12c05205beR39)? I think it would be possible for a MZ sign to have differing alert statuses for either line but be set to headway mode.

It looks like signs set to headways don't account for alert status at all. Am I reading this code wrong?

https://github.com/mbta/realtime_signs/blob/9f26a2267a681d20132f943481ef465f263f9299/lib/signs/utilities/headways.ex#L9-L39

PaulJKim:
In this [code path](https://github.com/mbta/realtime_signs/blob/main/lib/signs/utilities/messages.ex#L37) a mezzanine sign could be set to `:headway` which means that we first check the alert status. If the alert status happened to be in the shape of a tuple, that would result in `get_alert_messages` returning `nil`

lib/signs/realtime.ex
sloanelybutsurely: Yeah I suppose this is just preference on my part. This produces parallel structure across the two `case` clauses.

The structure being:
1. Get alert status(es)
2. Return alert status(es)

Which is pretty obvious for a single configuration

https://github.com/mbta/realtime_signs/blob/836b0f0cf0e53ba5a23e1fdcd65b93e0a96f0967/lib/signs/realtime.ex#L157-L159

But requires the additional variables for separate top and bottom configurations

https://github.com/mbta/realtime_signs/blob/836b0f0cf0e53ba5a23e1fdcd65b93e0a96f0967/lib/signs/realtime.ex#L151-L154

Removing the intermediate variable for a single configuration would remove the parallel structure:

```elixir
case sign.source_config do
  {top, bottom} ->
    top_alert_status = get_alert_status_for_source_config(top, sign.alerts_engine)
    bottom_alert_status = get_alert_status_for_source_config(bottom, sign.alerts_engine)

    {top_alert_status, bottom_alert_status}

  source_config ->
    get_alert_status_for_source_config(source_config, sign.alerts_engine)
end
```

---

<details>
<summary>(bonus content)</summary>

While on the subject, I think a way we would simplify _a lot_ of this code is with something like `SourceConfig.map/2` (and maybe `SourceConfig.simplify/1`) which would look something like this: https://github.com/mbta/realtime_signs/pull/798/files


</details>

PaulJKim (approved):
Looks good!

PR #800, opened 2024-07-22
fix: handle split alert statuses for signs in headway mode

sloanelybutsurely:
@PaulJKim pointed out that mezzanine signs when set to `:headway` mode check alert status but that code hadn't been updated to handle when alert statuses are different between `top` and `bottom` source configurations.

This change intercepts split alert status values passed to `get_alert_messages/2` and gets the highest priority alert status of the two using `Engine.Alerts.higher_priority_status/2` before continuing on to the existing implementation.

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -205,14 +205,24 @@ defmodule Signs.Utilities.Messages do
     end
   end
 
+  @spec get_alert_messages(
+          Engine.Alerts.Fetcher.stop_status()
+          | {Engine.Alerts.Fetcher.stop_status(), Engine.Alerts.Fetcher.stop_status()},
+          Signs.Realtime.t()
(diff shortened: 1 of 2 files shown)


sloanelybutsurely:
@PaulJKim I _think_ this should fix the issue you brought up here: https://github.com/mbta/realtime_signs/pull/797#issuecomment-2243073188

I'll work on adding a test case for this.

PaulJKim:
@sloanelybutsurely I'm not sure if we'd be maintaining the existing behavior by taking the first non-nil message. We might need to distill things back down to one 'max stop status' if it's a mezzanine sign with differing alert statuses for either line. https://github.com/mbta/realtime_signs/blob/main/lib/engine/alerts/fetcher.ex#L20-L52

PaulJKim (approved):
Looks good to me!

PR #810, opened 2024-08-26
Migrate from RTR to GTFS-rt

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Redeploy RTS Migration to Concentrate](https://app.asana.com/0/1185117109217422/1208135991889089/f)

This PR changes the predictions parsing logic so that we can read from the GTFS-rt feed instead of RTR.

This was previously attempted with #743 but after seeing some issues with `BRD` predictions and passthrough announcements, we had to rollback and do some investigation.

Following the investigation, some flaws in the filtering logic were uncovered so this new PR includes a minor adjustments which helps us more accurately filter out predictions where a train has already departed.

**Note**
There are still some things regarding passthrough announcements that need to be worked out before this PR can be merged which are described in [this ticket](https://app.asana.com/0/1185117109217422/1208135991889084/f)


.envrc.template
```diff
@@ -17,7 +17,6 @@ export API_V3_URL=https://api-dev-green.mbtace.com
 #export CHELSEA_BRIDGE_URL=
 #export CHELSEA_BRIDGE_AUTH=
 
-# URLs of the enhanced trip-update and vehicle-position feeds. Default to the real feed URLs if
-# not set here.
-#export TRIP_UPDATE_URL=
-#export VEHICLE_POSITIONS_URL=
+# URLs of the enhanced trip-update and vehicle-position feeds. Required.
```
sloanelybutsurely, line 20: This says "Required" but it looks like there are defaults provided in `config/runtime.exs`. Should we remove "Required" here or maybe `raise` in `config/runtime.exs` if these values are in the environment?

lib/predictions/predictions.ex
```diff
@@ -135,4 +153,10 @@ defmodule Predictions.Predictions do
       true
     end
   end
+
+  @spec has_departed?(Predictions.Prediction.t()) :: boolean()
+  defp has_departed?(prediction) do
+    prediction.seconds_until_departure && prediction.seconds_until_departure < 0 &&
+      not prediction.stopped_at_predicted_stop?
+  end
```
PaulJKim, line 161: Flagging for review: This is new logic since the first attempt at migrating

lib/predictions/predictions.ex
```diff
@@ -80,25 +83,27 @@ defmodule Predictions.Predictions do
          else: nil
 
     seconds_until_passthrough =
-      if stop_time_update["passthrough_time"],
-        do: stop_time_update["passthrough_time"] - current_time_seconds,
+      if not revenue_trip? or schedule_relationship == :skipped,
```
PaulJKim, line 86: Flagging for review: This check against `schedule_relationship` is new logic since the first attempt at migrating

sloanelybutsurely (approved):
code lgtm!

config/runtime.exs
```diff
@@ -12,12 +12,12 @@ if config_env() != :test do
...
       System.get_env(
         "TRIP_UPDATE_URL",
-        "https://s3.amazonaws.com/mbta-gtfs-s3/rtr/TripUpdates_enhanced.json"
+        "https://s3.amazonaws.com/mbta-gtfs-s3/concentrate/TripUpdates_enhanced.json"
       ),
     vehicle_positions_url:
       System.get_env(
         "VEHICLE_POSITIONS_URL",
-        "https://s3.amazonaws.com/mbta-gtfs-s3/rtr/VehiclePositions_enhanced.json"
+        "https://s3.amazonaws.com/mbta-gtfs-s3/concentrate/VehiclePositions_enhanced.json"
```
digitalcora, line 20: Just a drive-by note, these have "official" [public-facing URLs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs-realtime.md#enhanced-json-feeds) which are behind a CDN. We ended up using these in Screens LTOTD tracking rather than these internal S3 bucket URLs.

panentheos (approved):
Code looks good. We should make sure to give this some runtime in dev or dev-green before deploying if we haven't already.

config/runtime.exs
PaulJKim: Sounds good to me, thanks for calling that out. Does Screens LTOTD also use dev/dev-blue versions of the CDN URLs as well?

digitalcora: Only the production feeds have CDN URLs as far as I know — if we're fetching a non-prod feed those will probably have to still use S3 bucket URLs.

PR #817, opened 2024-09-12
formalize logging of sign content and generate tags

panentheos:
#### Summary of changes

**Asana Ticket:** [Improve RTS and SCU message logging](https://app.asana.com/0/1185117109217422/1208067121004443/f)

This adds `audio`, `visual`, and `tag` fields to RTS message logs, with JSON-parseable visual data.
* Splunk should be able to parse the visual page data, allowing us to transform it easily in dashboards. (We'll want to deploy to dev-green and test in splunk to make sure this works as expected.)
* Legacy audio message logs now contain a human-readable representation of what was played. Note that for legacy stations (all of them currently), this data isn't exactly what's played, but it should be semantically equivalent. After migration, these logs will show exactly what was played.
* The random `tag` is not helpful for legacy stations, but after migration, SCUs will log this tag in their own output, allowing us to trace message playback.

### DIFF
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -23,20 +23,26 @@ defmodule PaEss.Updater do
         _ -> nil
       end
 
-    log_meta = [sign_id: id, current_config: log_config]
+    visual = zip_pages(top, bottom) |> format_pages()
+    tag = create_tag()
+
+    log_meta = [
+      sign_id: id,
+      current_config: log_config,
+      visual: Jason.encode!(visual),
+      tag: inspect(tag)
+    ]
 
     if config_engine.scu_migrated?(scu_id) do
-      pages = zip_pages(top, bottom)
-
       PaEss.ScuQueue.enqueue_message(
         scu_id,
         {:background, scu_id,
          %{
            visual_zones: [text_zone],
-           visual_data: format_pages(pages),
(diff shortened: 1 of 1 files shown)


cmaddox5 (approved):
Looks good!

lib/pa_ess/updater.ex
```diff
@@ -23,20 +23,26 @@ defmodule PaEss.Updater do
         _ -> nil
       end
 
-    log_meta = [sign_id: id, current_config: log_config]
+    visual = zip_pages(top, bottom) |> format_pages()
+    tag = create_tag()
```
cmaddox5, line 27: Actually, a question came to mind while working on the dashboard. Is it possible to default `tag` to `nil` for legacy? It would make it easier to tell legacy and migrated messages apart and would allow me to display `N/A` in the playback column.

panentheos: Instead of that, how about another explicit field that specifies whether it's legacy? Part of the goal here is to start unifying the shape of the logs as much as possible, so I wanted to start exposing this data so we can build dashboards to that spec (but yeah, we still need a way to differentiate legacy messages)

cmaddox5: Yea I like that better. Much less likely to cause confusion way down the line.

PR #818, opened 2024-09-19
streamline message logging

panentheos:
#### Summary of changes

**Asana Ticket:** [Improve RTS and SCU message logging](https://app.asana.com/0/1185117109217413/1208067121004443/f)

* Switches to explicit logging of extra message metadata, instead of dumping the raw struct. This effectively removes a lot of fields that were initially intended for debugging, but now that we're including visual and audio strings with every message, most of them should be unnecessary. This approach will let us add additional metadata easily, if desired.
* Keeps the PA message metadata needed for the playback dashboard, but renames them `pa_message_id`, `pa_message_priority`, and `pa_message_interval` to avoid confusion
* Unifies the PA message playback code, for simplicity and to allow bus sign logs to inherit the same rich log output.

### DIFF
--- a/lib/content/audio.ex
+++ b/lib/content/audio.ex
@@ -22,4 +22,6 @@ defprotocol Content.Audio do
   def to_params(audio)
   @spec to_tts(Content.Audio.t()) :: tts_value()
   def to_tts(audio)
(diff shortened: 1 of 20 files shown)


digitalcora (approved):
Nice improvement!

PR #819, opened 2024-09-26
Don't deploy to dev-green on push to main

PaulJKim:
#### Summary of changes

**Asana Ticket:** [[extra] remove auto-deploy for RTS dev-green](https://app.asana.com/0/1185117109217413/1208381515035004/f)

Automatic deploys to dev-green upon pushing to main can cause unintentional hindrances when using dev-green to test branches. This change to the workflow makes it so that we can only manually deploy to dev-green.

### DIFF
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -1,9 +1,6 @@
 name: Deploy to Dev Green
 
-on:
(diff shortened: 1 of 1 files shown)


digitalcora (approved):
🙅‍♀️ 🔄 ✅

PR #822, opened 2024-09-30
simplify headway message code

panentheos:
#### Summary of changes

This is a small refactor in advance of a larger one. It cleans up some unused abstraction around vehicle types in the headway message code, since we only ever generate messages about trains. Also flips the display logic around so the `destination` field controls which primary code path is used, rather than `routes`. This should result in no user-facing changes.

Note: Regarding the recent discussion about adding SL headways, the bus logic doesn't use this code, so that doesn't impact the decision to clean up this code.

### DIFF
--- a/lib/content/message/headways/top.ex
+++ b/lib/content/message/headways/top.ex
@@ -1,68 +1,26 @@
 defmodule Content.Message.Headways.Top do
-  require Logger
-  defstruct [:destination, :vehicle_type, :routes]
+  defstruct [:destination, :routes]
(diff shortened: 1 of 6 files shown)


cmaddox5 (approved):
Makes sense. Will part of the SL headway task be to allow that part of the bus code to look in this module? Or would we create a new module for that?

panentheos:
The bus doesn't hook into the `Content.Message` framework at all, so those changes will be fully separate.

PR #823, opened 2024-10-01
pass single route through messages

panentheos:
#### Summary of changes

This is a small refactor in advance of a larger one. Several of the `Message` structs carry a `routes` field which holds the list of routes from the config used to generate it. However, at the end of the computation, the only thing we care about is whether there was a single unique route. This changes the fields to hold a single `route`, or nil if there were multiple, which simplifies downstream calculations. It also adds a `route` field to a few additional messages, which will be used by later changes.


PaulJKim (approved):
Looks good

PR #827, opened 2024-10-07
streamline prediction destination code

panentheos:
#### Summary of changes

This change is in support of an larger refactor. It removes some soft error handling in `destination_for_prediction`, simplifying the callers. This code has been stable for a long time, and the logs show no indication that it ever fails during normal operation. Going forward, any unexpected predictions would cause this to fail loudly, which seems like a decent approach.

### DIFF
--- a/lib/content/message/predictions.ex
+++ b/lib/content/message/predictions.ex
@@ -88,34 +88,23 @@ defmodule Content.Message.Predictions do
         do: {nil, nil},
         else: do_crowding(prediction, sign)
 
-    case Content.Utilities.destination_for_prediction(
-           prediction.route_id,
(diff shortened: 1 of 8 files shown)


digitalcora (approved):
Let It Crash™

Just to be clear about the failure boundaries, if this were to ever crash (and it wasn't a transient condition that resolved on process restart), what would happen with signs? Would the issue be limited only to specific signs?

panentheos:
We've tuned the supervision layer so that individual sign crashes don't bring down the entire system, they just continuously restart (with a small delay). So the impact would be limited to signs seeing unexpected data, and they would recover automatically when the data went away.

PR #831, opened 2024-10-16
move terminal flag to source config object

panentheos:
#### Summary of changes

This moves the `terminal` flag from the individual source object to the config object. This is a more natural way to model this information, and all instances of this flag were identical across all sources within each config. This simplifies lookup of this information during the main logic loop.

The `signs.json` transformation was done programmatically, and the script is included for review.

Also includes a small unrelated script that performs some zone mapping analysis.


cmaddox5 (approved):
Looks good. I agree, this does make more sense.

PR #832, opened 2024-10-17
move platform config out of signs.json

panentheos:
#### Summary of changes

This moves the `platform` field out of the sign configuration and into code. The platform is only non-nil in two specific cases, and can be straightforwardly computed from the stop id, so there's no reason to store it separately.

The `signs.json` change was made programmatically, and the script is included for review.


cmaddox5 (approved):
Looks good!

PR #834, opened 2024-10-25
fix capitalization of early AM messages

panentheos:
#### Summary of changes

**Asana Ticket:** [Lower case "g" in "Oak grove due" should be capitalized](https://app.asana.com/0/1185117109217422/1208591545042708/f)

`String.capitalize` lowercases the remainder of the string, which is usually not what we want. These usages are unnecessary anyway, since the destination names already have the right casing.


digitalcora (approved):
✂️ ✨

PR #835, opened 2024-10-25
don't modify predictions when both lines ARR

panentheos:
#### Summary of changes

This is a small behavior change that was ok'd by product. There was some old logic that would detect a case where `ARR` would be shown on both lines, and tweak the display so the second line said `1 min` in all but a handful of cases. We're not exactly sure of the original rationale, but suspect it was to smooth over a data quality issue. The data should be better now, and also this special handling creates an inconsistency with other touchpoints, so we'd like to remove it.

### DIFF
--- a/test/signs/realtime_test.exs
+++ b/test/signs/realtime_test.exs
@@ -540,7 +540,7 @@ defmodule Signs.RealtimeTest do
       Signs.Realtime.handle_info(:run_loop, @sign)
     end
 
-    test "does not allow ARR on second line if platform does not have multiple berths" do
+    test "allows ARR on second line" do
       expect(Engine.Predictions.Mock, :for_stop, fn _, _ ->
         [
           prediction(destination: :cleveland_circle, arrival: 15, stop_id: "1"),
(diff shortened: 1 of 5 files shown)


test/signs/realtime_test.exs
```diff
@@ -540,15 +540,15 @@ defmodule Signs.RealtimeTest do
       Signs.Realtime.handle_info(:run_loop, @sign)
     end
 
-    test "does not allow ARR on second line if platform does not have multiple berths" do
+    test "allows ARR on second line" do
```
digitalcora, line 543: Is there value in keeping this as a distinct test case? I suspect if we were writing this from scratch and this "multi-berth" feature had never existed, we wouldn't think to have it, since it's not covering any unique code path. (Though maybe we'd want to test that ordering works correctly when the displayed time is "ARR" for both entries.)

digitalcora (approved):

test/signs/realtime_test.exs
panentheos: I thought about that too. Obviously it's nice to see the delta to confirm that this change works, but there may not be a ton of value beyond that. I'm inclined to leave it for now, but it could be a candidate for sweeping up down the road, certainly.

PR #836, opened 2024-10-28
reference prediction directly and remove individual fields

panentheos:
#### Summary of changes

Prediction and stopped-train messages, unsurprisingly, end up needing most of the fields from the underlying prediction. This stores a reference to the prediction struct itself, rather than copying each individual field, which simplifies the representation and reduces boilerplate. Also inlines the `width` parameter, which never actually varies.


cmaddox5 (approved):
Nice cleanup. Looks good.

PR #837, opened 2024-10-28
use day-of-week based headway periods

panentheos:
#### Summary of changes

**Asana Ticket:** [Change RTS to read these new values, handle appropriately](https://app.asana.com/0/1185117109217413/1208512927678681/f)

This switches to reading the new day-of-week based headway periods.

Also modifies the config parsing code so it doesn't log errors when encountering unexpected data (which we'll have during the transition). Note that the current code _will_ log errors in this case, so we expect to see some errors logged between deploying the signs-ui change and deploying this.

### DIFF
--- a/lib/engine/config/headways.ex
+++ b/lib/engine/config/headways.ex
@@ -31,20 +31,16 @@ defmodule Engine.Config.Headways do
 
   @spec parse(map()) :: [Headway.t()]
   def parse(map) do
-    Enum.flat_map(map, fn {group, group_conf} ->
-      Enum.flat_map(group_conf, fn {time_period, headway_conf} ->
-        case Headway.from_map(group, time_period, headway_conf) do
-          {:ok, conf} ->
-            [conf]
-
-          :error ->
-            Logger.error(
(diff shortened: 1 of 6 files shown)


lib/engine/config/headways.ex
```diff
@@ -31,20 +31,16 @@ defmodule Engine.Config.Headways do
...
-              "event=group_headways_parse_error headway_id=#{inspect({group, time_period})} group_conf=#{inspect(group_conf)}"
-            )
-
-            []
-        end
-      end)
-    end)
+    for {group, periods} <- map,
+        {period, %{"range_low" => low, "range_high" => high}} <- periods,
+        time_period = parse_time_period(period) do
```
PaulJKim, line 36: since `time_period` could be `nil` and then is passed to the `Headway` struct, shouldn't the typespec for `Headway.time_period()` also include `| nil`?

panentheos: We're relying on the fact that falsy values act like a filter, so they actually get skipped and we don't create `Headway` structs for those.

panentheos:
Closing in favor of #839

PR #838, opened 2024-10-28
defer crowding and new cars calculation until it's needed

panentheos:
#### Summary of changes

This moves the crowding and "new cars" calculations out of the prediction message code, with the goal of further simplifying that code. Those calculations only affect audio announcements, but they were being performed earlier in the pipeline because that was the most convenient place that could access both the sign and prediction structs. Now that the prediction struct is being fed through directly, it's possible to perform the crowding and new-cars calculations right where they're actually needed.

This should have no user-facing changes, and the test diffs only show fewer internal location lookups due to the deferred processing. Note that actual crowding _output_ is still disabled in code, so that will be suppressed regardless.


digitalcora (approved):
It seems like there's enough logic in `crowding_description` and that collection of private functions for it to live in its own module, perhaps in a namespace that makes it clearer where it's used / what its purpose is. Other than that thought, looks nice!

PR #842, opened 2024-11-01
unify prediction message constructors

panentheos:
#### Summary of changes

This refactors the creation of prediction messages to unify the `terminal` and `non_terminal` code paths, since they're more similar than different. It also introduces a new `special_sign` field to encode various special case behaviors without having to check against station code/zone repeatedly.

### DIFF
--- a/lib/signs/utilities/audio.ex
+++ b/lib/signs/utilities/audio.ex
@@ -321,7 +321,6 @@ defmodule Signs.Utilities.Audio do
              bottom.__struct__ in [Message.Predictions, Message.StoppedTrain] ->
         [{:predictions, [top, bottom]}]
 
-      # This only occurs at the RJFK mezzanine zone
       {%Message.Predictions{}, %Message.PlatformPredictionBottom{}} ->
         [{:predictions, [%{top | special_sign: :jfk_mezzanine}]}]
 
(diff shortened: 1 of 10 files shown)


lib/signs/utilities/audio.ex
```diff
@@ -323,7 +323,7 @@ defmodule Signs.Utilities.Audio do
 
       # This only occurs at the RJFK mezzanine zone
```
digitalcora, line 324: Maybe this comment isn't needed anymore? Since the special case is explicit in the pattern (`special_sign: :jfk_mezzanine`).

lib/signs/utilities/predictions.ex
```diff
@@ -46,18 +42,19 @@ defmodule Signs.Utilities.Predictions do
     end)
     |> filter_large_red_line_gaps()
     |> Enum.map(fn prediction ->
-      cond do
-        stopped_train?(prediction) ->
-          Content.Message.StoppedTrain.from_prediction(prediction)
-
-        terminal? ->
-          Content.Message.Predictions.terminal(prediction, station_code, zone)
+      special_sign =
```
digitalcora, line 45: Since this is only used in the `else` branch, should we move it there, to make it clear that this logic doesn't affect the `StoppedTrain` case?

digitalcora (approved):
Nice improvement! A couple non-blocking comments.

PR #844, opened 2024-11-04
unify no service audio code paths

panentheos:
#### Summary of changes

This refactors the `Closure` and `NoServiceToDestination` audio modules into a single `NoService` audio. These code paths were closely related to each other, and unifying them simplifies the calling code. Test coverage is maintained by the suite of higher level cases in `realtime_test.exs`. Also includes a few minor simplifications to the headway and first-train audio constructors.


test/content/audio/vehicles_to_destination_test.exs
```diff
@@ -32,138 +32,17 @@ defmodule Content.Audio.VehiclesToDestinationTest do
...
-                 destination: nil,
-                 headway_range: {8, 10}
-               }
-             ] =
-               from_headway_message(
-                 %Content.Message.Headways.Top{destination: nil},
-                 %Content.Message.Headways.Bottom{range: {8, 10}}
-               )
-    end
+  test "from_messags/2" do
```
PaulJKim, line 35: small typo here

lib/content/audio/no_service.ex
```diff
@@ -0,0 +1,105 @@
...
+           use_routes?: use_routes?
+         }) do
+      shuttle = if(use_shuttle?, do: " Use shuttle.", else: "")
+
+      cond do
+        destination ->
+          {:ok, destination_text} = PaEss.Utilities.destination_to_ad_hoc_string(destination)
+          "No #{destination_text} service.#{shuttle}"
+
+        use_routes? ->
```
PaulJKim, line 95: should this clause not come first in the `cond` block? Based on the definition for `from_messages/2` above for this special case, it looks like `destination` could still be truthy

panentheos: Currently it doesn't matter, because we always generate a destination-less `Message.NoService` for this case. I've got a followup PR that changes that code path and this logic, but I went ahead and swapped it anyway.

PR #852, opened 2024-12-03
remove unused fallbacks for destination utils

panentheos:
#### Summary of changes

This removes some fallback code dealing with destination names, which is never actually needed in practice, because we can produce content for all supported destinations. This simplifies the calling code. Adds `destination_var` values for Chelsea and South Station for completeness (though they are not strictly required, since we don't use those destinations from the rail code, and the bus code has its own way of handling that).

### DIFF
--- a/lib/content/audio/stopped_train.ex
+++ b/lib/content/audio/stopped_train.ex
@@ -41,15 +41,13 @@ defmodule Content.Audio.StoppedTrain do
     @stopped "641"
 
     def to_params(audio) do
-      dest_var = PaEss.Utilities.destination_var(audio.destination)
-
       if Utilities.directional_destination?(audio.destination) do
         do_ad_hoc_message(audio)
       else
         vars = [
           @the_next,
           @train_to,
-          dest_var,
+          PaEss.Utilities.destination_var(audio.destination),
           @is,
           @stopped,
           number_var(audio.stops_away),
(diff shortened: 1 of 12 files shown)


lib/content/audio/stopped_train.ex
```diff
@@ -41,26 +41,22 @@ defmodule Content.Audio.StoppedTrain do
...
-              number_var(audio.stops_away),
-              stops_away_var(audio.stops_away)
-            ]
-
-            PaEss.Utilities.take_message(vars, :audio)
-          end
-
-        {:error, :unknown} ->
-          do_ad_hoc_message(audio)
+      dest_var = PaEss.Utilities.destination_var(audio.destination)
```
digitalcora, line 44: This is only used in the `else` branch below, so it could live there instead (and only once, so maybe no need for it to even be a variable).

lib/content/audio/train_is_boarding.ex
```diff
@@ -51,16 +51,12 @@ defmodule Content.Audio.TrainIsBoarding do
...
-        {:ok, destination_var} ->
-          if PaEss.Utilities.directional_destination?(audio.destination) do
-            do_ad_hoc_message(audio)
-          else
-            do_to_params(audio, destination_var)
-          end
-
-        {:error, :unknown} ->
-          do_ad_hoc_message(audio)
+      destination_var = PaEss.Utilities.destination_var(audio.destination)
```
digitalcora, line 54: Same comment as previous.

digitalcora (approved):
🧹

PR #859, opened 2024-12-16
Fix harvard upper busway stop ids

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Fix up Harvard upper busway stops (third time's the charm!)](https://app.asana.com/0/1185117109217413/1208983261541535/f)

Fix stop ids after BNR initial launch

### DIFF
--- a/priv/signs.json
+++ b/priv/signs.json
@@ -9234,7 +9234,7 @@
       {
         "sources": [
           {
-            "stop_id": "76123",
+            "stop_id": "76124",
             "route_id": "71",
             "direction_id": 0
           }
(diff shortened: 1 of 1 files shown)


PaulJKim:
@robbie-sundstrom nope no need for a signs UI pr! Once this goes out, we'll start sending the missing predictions to the harvard busway sign right away. Signs UI just acts as a viewer for us to visualize what we're sending out to signs, along with some controls that OIOs use to set signs into different modes and configure headway values etc.

digitalcora:
@PaulJKim It _is_ part of the PR checklist that changes to `signs.json` here should also be propagated to SignsUI. Maybe there are some categories of change to `signs.json` that don't have any effect on the way SignsUI uses the file, but I'm not sure off-hand if this is one of those. Do you have a better handle on this?

PaulJKim:
Well actually, we don't include the busway signs in the stop_id to sign mapping so we don't need to update the RTS dependency here since it won't affect signs UI

PR #860, opened 2024-12-17
standardize train description language

panentheos:
#### Summary of changes

**Asana Ticket:** ["Train to Ashmont" vs. "Ashmont train" discrepancy with track numbers](https://app.asana.com/0/1185117109217422/1208651243993392/f)

This converts various audio messages to use the standard train description language, specifically: `[destination] train`, except for GL, which uses `[branch] train to [destination]`.

The change is divided into two commits for easier review, with the first one containing the actual behavior change. As part of this, we switch from using predefined canned messages (which are mapped to take ids on ARINC's side) to directly specifying the takes, which makes for a more legible approach. This results in some very different-looking output, but since we're using the same underlying takes, the actual audio readouts should be unchanged (except where we're deliberate changing the train description).

The impacted message ids are:
* 141: The next ___ train ___ in one minute
* 90: The next ___ train ___ in ___ minutes
* 142: The next ___ train on the ___ platform ___ in one minute
* 98: The next ___ train ___ in ___ minutes on the ___ platform
* 159: The following ___ train ___ in one minute
* 160: The following ___ train ___ in ___ minutes

The second commit introduces a small helper that eliminates some repetition in the tests and dramatically shrinks the line count as a result. This is a purely mechanical change.


lib/content/audio/following_train.ex
```diff
@@ -48,31 +48,14 @@ defmodule Content.Audio.FollowingTrain do
...
-      else
-        green_line_branch = Content.Utilities.route_branch_letter(audio.route_id)
-
-        cond do
-          !is_nil(green_line_branch) ->
-            green_line_with_branch_params(audio, green_line_branch, dest_var)
-
-          audio.minutes == 1 ->
-            {:canned, {"159", [dest_var, verb_var(audio)], :audio}}
+      min_or_mins = if(audio.minutes == 1, do: :minute, else: :minutes)
```
robbie-sundstrom, line 52: I am still learning, so I'm not 100% sure if this is correct or necessary, but could the same thing be achieved more functionally with pattern matching like so:
```  
def min_or_mins(%{minutes: 1}), do: :minute

def min_or_mins(%{minutes: _}), do: :minutes
  ```
  
 and then it could be called with `min_or mins(audio)` here and later in this file.

Comment (panentheos) [reply lib/content/audio/following_train.ex:52]: Yes, that would work, although the code below is slightly different because it returns strings instead of atoms, so it's not quite reusable in this case.

More generally, I think pattern matching can be a great tool in a lot of situations, but it depends on the context what the clearest option is. I picked `if` here because there are just two branches, and the logic is nice and compact. `case` really shines when there are multiple branches that are logically related. Multi-headed functions like that tend to work well when the branches are more logically disjoint. And of course re-usability plays a part too.

Comment (robbie-sundstrom) [reply lib/content/audio/following_train.ex:52]: Gotcha, thanks for explaining your thought process here! I understand how `if` makes the most sense here then

File: test/content/audio/train_is_boarding_test.exs [hunk 2]
```diff
@@ -48,7 +36,7 @@ defmodule Content.Audio.TrainIsBoardingTest do
 
       assert Content.Audio.to_params(audio) ==
                {:canned,
-                {"111", ["501", "21000", "507", "21000", "4016", "21000", "544", "21000", "542"],
+                {"111", ["501", "21000", "4016", "21000", "864", "21000", "544", "21000", "542"],
```
robbie-sundstrom, line 39: Might be a nitpick, but could you do something here like in `realtime_test.exs` with the `expect_audios` function and check the `tts_value` of the `Audio`? It's difficult to parse what's changing in all of these tests when converting between the numbers and the corresponding text value.

panentheos: Not in this particular case, because this is testing `to_params`, which deals just with the numerical take ids. I agree it's a pain, but it's sort of an inherent pain with the current system. The way I've been verifying it is by searching the code for the take id, which pretty consistently gets you to a line that defines the mapping, so you know what it ends up saying. In this case, `"507" -> "train to", "4016" -> "Ashmont"` becomes `"4016" -> "Ashmont", "864" -> "train"`, which is the purpose of this change.

test/signs/realtime_test.exs
```diff
@@ -533,9 +515,7 @@ defmodule Signs.RealtimeTest do
       expect_messages({"Mattapan       BRD", "Mattapan     2 min"})
 
       expect_audios(
-        [
-          {:canned, {"109", ["501", "21000", "507", "21000", "4100", "21000", "544"], :audio}}
-        ],
+        [{:canned, {"109", ["501", "21000", "4100", "21000", "864", "21000", "544"], :audio}}],
```
PaulJKim, line 518: couple spots like this where we could use the new `spaced/1` helper

test/content/audio/following_train_test.exs
```diff
@@ -12,22 +12,7 @@ defmodule Content.Audio.FollowingTrainTest do
...
-                   "21000",
-                   "503",
-                   "21000",
-                   "504",
-                   "21000",
-                   "5005",
-                   "21000",
-                   "505"
-                 ], :audio}}
+                {"115", spaced(["667", "4016", "864", "503", "504", "5005", "505"]), :audio}}
```
digitalcora, line 15: Suggestion: We could make these lists more concise still by expressing them as

```elixir
~w[667 4016 864 503 504 5005 505]
```

robbie-sundstrom (approved):

test/content/audio/following_train_test.exs
panentheos: Nice idea, will probably handle that in a followup PR

PR #863, opened 2024-12-21
Migrate to update_type

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Migrate to Update_type for trip type](https://app.asana.com/0/1185117109217413/1206604799720436/f)

RTS currently uses the `uncertainty` field in order to distinguish different types of predictions (mid trip, terminal, and reverse). This field will eventually be deprecated by the Transit Data team and has been replaced by a new field provided at the trip level called `update_type`. This PR migrates RTS to using this new field in favor of `uncertainty` which also made it easier to clean up the predictions feed parsing code a bit as well.

#### Reviewer Checklist

### DIFF
--- a/lib/predictions/predictions.ex
+++ b/lib/predictions/predictions.ex
@@ -2,6 +2,8 @@ defmodule Predictions.Predictions do
   alias Predictions.Prediction
   require Logger
 
+  @excluded_prediction_types [:reverse]
+
   @spec get_all(map(), DateTime.t()) ::
           {%{
              optional({String.t(), integer()}) => [Predictions.Prediction.t()]
@@ -10,22 +12,8 @@ defmodule Predictions.Predictions do
     predictions =
       feed_message["entity"]
       |> Stream.map(& &1["trip_update"])
-      |> Stream.filter(
-        &(relevant_rail_route?(&1["trip"]["route_id"]) and
-            &1["trip"]["schedule_relationship"] != "CANCELED")
-      )
-      |> Stream.flat_map(&transform_stop_time_updates/1)
-      |> Stream.filter(fn {update, _, _, _, _, _, _} ->
-        (update["arrival"] && update["arrival"]["uncertainty"]) ||
-          (update["departure"] && update["departure"]["uncertainty"]) ||
-          update["passthrough_time"]
-      end)
-      |> Stream.map(&prediction_from_update(&1, current_time))
-      |> Enum.reject(
-        &((is_nil(&1.seconds_until_arrival) and is_nil(&1.seconds_until_departure) and
-             is_nil(&1.seconds_until_passthrough)) or
-            has_departed?(&1))
-      )
+      |> Stream.filter(&valid_trip_update?/1)
+      |> Stream.flat_map(&trip_update_to_predictions(&1, current_time))
 
     vehicles_running_revenue_trips =
       predictions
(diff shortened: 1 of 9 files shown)


test/predictions/predictions_test.exs
```diff
@@ -372,7 +368,7 @@ defmodule Predictions.PredictionsTest do
               }, _} = get_all(feed_message, @current_time)
     end
 
-    test "include predictions with low uncertainty" do
+    test "include non reverse predictions" do
       reassign_env(:filter_uncertain_predictions?, true)
```
digitalcora, line 372: drive-by: Could we rename this flag, since we're otherwise removing the concept of "uncertainty" from the code? Maybe `filter_reverse_predictions?`. (Curious why this is configurable in the first place.)

PaulJKim: Yeah, we should definitely rename it but I was thinking maybe in a small follow-up task or something since we'd want to change that in the devops repo as well. In terms of why it's configurable, I'd guess it's because back then RTS was running on opstech3 and we would often take advantage of flags like this to quickly change up flags to adjust what RTS was doing. It was added in a really old PR a few years back and may not be necessary anymore. https://github.com/mbta/realtime_signs/pull/369

digitalcora: Hm, yeah, maybe more useful initially as a "feature flag" in the time immediately after the feature was added. I think we could outright remove it now, if @panentheos agrees.

PaulJKim: Yeah you're probably right. The feature flag isn't something we'd likely prefer to use in a situation where we want to filter out any sort of prediction type at large anyway since that requires infra to get involved these days. In the more recent past when we've filtered out terminal/reverse predictions on some heavy rail lines, we just introduced a temporary code change and then revert it when predictions were trustworthy again.

lib/predictions/predictions.ex
```diff
@@ -38,77 +26,98 @@ defmodule Predictions.Predictions do
...
+  @spec trip_update_to_predictions(map(), DateTime.t()) :: [Prediction.t()]
+  defp trip_update_to_predictions(trip_update, current_time) do
     current_time_seconds = DateTime.to_unix(current_time)
+    destination_stop_id = get_destination_stop_id(trip_update)
+    vehicle_id = trip_update["vehicle"]["id"]
+    vehicle_location = Engine.Locations.for_vehicle(vehicle_id)
+    route_id = trip_update["trip"]["route_id"]
+    direction_id = trip_update["trip"]["direction_id"]
+    trip_id = trip_update["trip"]["trip_id"]
+    revenue_trip? = trip_update["trip"]["revenue"]
```
panentheos, line 43: nit: We could inline some of these below to shorten the boilerplate a bit. Alternatively, we could destructure the function param directly, as long as these fields will always be present.

lib/predictions/predictions.ex
```diff
@@ -142,23 +151,41 @@ defmodule Predictions.Predictions do
...
+    if Application.get_env(:realtime_signs, :filter_uncertain_predictions?),
+      do: prediction.type in @excluded_prediction_types,
+      else: false
+  end
+
+  @spec has_departed?(Prediction.t()) :: boolean()
   defp has_departed?(prediction) do
-    prediction.seconds_until_departure && prediction.seconds_until_departure < 0 &&
-      not prediction.stopped_at_predicted_stop?
+    prediction.seconds_until_departure < 0 and not prediction.stopped_at_predicted_stop?
```
panentheos, line 189: nit: Technically this works, because `nil < 0 == false` according to Elixir's term ordering rules, but I think I prefer the explicit check as a reminder that `seconds_until_departure` might be `nil`

test/predictions/predictions_test.exs
panentheos: Yep, I don't think we need as a configurable option. Let's remove it!

test/signs/realtime_test.exs
```diff
@@ -1928,7 +1926,8 @@ defmodule Signs.RealtimeTest do
       stopped_at_predicted_stop?: Keyword.get(opts, :stopped_at_predicted_stop, false),
       boarding_status: Keyword.get(opts, :boarding_status),
       revenue_trip?: true,
-      vehicle_id: "v1"
+      vehicle_id: "v1",
+      type: Keyword.get(opts, :prediction_type, :mid_trip)
```
panentheos, line 1930: nit: Why not name this option `type` for consistency?

panentheos (approved):
See some small comments, otherwise looks good!

PR #864, opened 2024-12-30
Restructure alerts engine to merge new results into cache instead of delete/write

robbie-sundstrom:
#### Summary of changes

**Asana Ticket:** [Refactor RTS Alerts Engine](https://app.asana.com/0/1185117109217413/1208885845442437)

- On every update to the alerts written to ETS, instead of deleting all and writing all of the new alerts, we now replace them all
- Added an ETS engine file to handle the `replace_all` method for reusability

#### Reviewer Checklist


lib/engine/ets.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Engine.ETS do
```
robbie-sundstrom, line 1: Also thought about naming this module `Engine.Cache` since ETS is acting as our cache.

lib/engine/ets.ex
```diff
@@ -0,0 +1,34 @@
...
+    new_keys = MapSet.new(new_entries, &elem(&1, 0))
+    current_table_keys = keys(table)
+
+    removed_keys = MapSet.difference(current_table_keys, new_keys)
+    Enum.each(removed_keys, &:ets.delete(table, &1))
+
+    :ets.insert(table, Enum.into(new_entries, []))
+  end
+
+  # Returns a MapSet of all keys in the table.
```
PaulJKim, line 24: I think we could achieve the same thing by doing

```
keys = 
  table
  |> :ets.tab2list()
  |> Enum.map(fn {key, _value} -> key end)
  |> MapSet.new()
  ```

File: lib/engine/ets.ex (hunk 1 above)
Comment (PaulJKim) [reply lib/engine/ets.ex:1]: Hmm yeah interesting thought - it's true that ETS essentially acts as our cache, although I'm not sure how I feel about abstracting it as such. It could help the cleanliness of some of the engine logic if we abstracted away the direct calls to `:ets` I suppose. Curious to see if @panentheos has thoughts

Comment (PaulJKim) [reply lib/engine/ets.ex:1]: Hmm yeah interesting thought - it's true that ETS essentially acts as our cache, although I'm not sure how I feel about abstracting it as such. It could help the cleanliness of some of the engine logic if we abstracted away the direct calls to :ets I suppose. Curious to see if @panentheos has thoughts

**Comment by panentheos** [comment]
While the delete/insert sequence in the original code is indeed error-prone, ETS does guarantee atomicity for several single operations, including bulk inserts. We leverage that to avoid the problem elsewhere by defining a sentinel "empty" value, overwriting all old values with that, merging the new values, and then writing the whole thing back in one operation. There's been a bit of consolidation of the code for this, and this would be a good opportunity to finish that up. See `lib/engine/locations.ex` and `lib/engine/predictions.ex`, which have essentially identical implementations, which could be pulled out as a shared utility function.

File: lib/engine/ets.ex (hunk 1 above)
Comment (panentheos) [reply lib/engine/ets.ex:1]: I agree with Paul, I'm not sure there's much to be gained from adding new terminology here. Also, we usually use the name `Engine` to describe GenServer processes that are actively doing something, like polling data repeatedly. Since these are just functions, I'd probably file them under something with "Utilities" in the name, or just add them to one of the existing utility modules if they fit there.

Comment (robbie-sundstrom) [reply lib/engine/ets.ex:1]: Gotcha, that distinction for engines makes sense to me! Given the other common use cases you pointed out, I think having this as an `ETSUtilities` module works well

**Comment by robbie-sundstrom** [comment]
@panentheos I did pull out the implementations into a shared utility module, but the behavior of the predictions/locations ETS table keys is different than what we want for alerts. Because of that, I kept them as two separate utility functions. I thought about combining them into one with two methods based on the number of inputs, but I ultimately decided that ended up being more complex.

**Review by panentheos — COMMENTED** [review]
This is looking good. See comments for additional recommendations.

File: lib/signs/utilities/ets_utils.ex [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+    :ets.tab2list(table)
+    |> Enum.map(&{elem(&1, 0), empty_value})
+    |> Map.new()
+    |> Map.merge(values)
+    |> Map.to_list()
+    |> then(&:ets.insert(table, &1))
+  end
+
+  @doc "Writes new key-values to table and removes any existing keys from ETS that aren't in new_entries."
+  def replace_contents(table, new_entry) when is_tuple(new_entry) do
```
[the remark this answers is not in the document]
robbie-sundstrom: Makes sense -- I was thinking of saving space in RTS by deleting old keys, but I don't think that will be a large issue since most of them are re-used. Updated the code with this suggestion.

test/signs/utilities/ets_utils_test.exs
```diff
@@ -0,0 +1,124 @@
...
+
+  setup do
+    # Create a fresh table before each test
+    table = create_table()
+    {:ok, table: table}
+  end
+
+  ###################################################################
+  # write_ets tests
+  ###################################################################
```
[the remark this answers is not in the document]
digitalcora: See `ExUnit.Case.describe/2` for documentation on that! Also, when using this to group tests for different functions of a single module, it's conventional to use the function's name as the message. For example

```elixir
describe "write_ets/3" do
  test "does the thing" do
    # ...
  end
end
```

These are available unqualified in test modules because of [`use ExUnit.Case`](https://github.com/elixir-lang/elixir/blob/v1.17.2/lib/ex_unit/lib/ex_unit/case.ex#L283-L294).

lib/signs/utilities/ets_utils.ex
```diff
@@ -0,0 +1,26 @@
...
+    |> Map.to_list()
+    |> then(&:ets.insert(table, &1))
+  end
+
+  @spec write_ets(:ets.tab(), [{any(), any()}], any()) :: boolean()
+  def write_ets(table, values_list, empty_value) when is_list(values_list) do
+    write_ets(table, Map.new(values_list), empty_value)
+  end
+
+  @spec write_ets(:ets.tab(), {any(), any()}, any()) :: boolean()
```
panentheos, line 22: Note: multiple specs are fine, and they will be automatically unioned together to form the final spec. Another option would be to have a single spec, and use the `|` operator to explicitly create the union for the second param.

panentheos (approved):
Looks good!

lib/signs/utilities/ets_utils.ex
robbie-sundstrom: I actually liked the separate specs over one unioned spec. I thought it was more quickly readable vs. the longer spec which takes a bit to digest.

digitalcora: > multiple specs are fine, and they will be automatically unioned together to form the final spec

I don't think this is quite true — it's not relevant to this situation, but Dialyzer does support dependent typing of return values (only with explicit specs, not inferred ones). In other words:

```elixir
@spec foo(integer()) :: :ok
@spec foo(binary()) :: :ok | :error
```

A call to `foo(17)` will never complain that you're not matching the `:error` case, but a call to `foo("bar")` might. See [Learn You Some Erlang](https://learnyousomeerlang.com/dialyzer) towards the end of the section **Typing Functions** (search _"fancier"_).

PR #867, opened 2025-01-07
extract prediction helpers

panentheos:
#### Summary of changes

This extracts several helpers related to predictions, as a part of the larger message refactoring effort. Notes:

* Removes special handling of "approaching" predictions, since only the audio path cares about them.
* Unifies the logic for determining "approximate" minute values, fixing some potential edge cases.
* Fixes a rare historical edge case where approaching predictions wouldn't be read out during the normal read loop.

Aside from these few cases, the overall system behavior should be unchanged.

### DIFF
--- a/lib/pa_ess/utilities.ex
+++ b/lib/pa_ess/utilities.ex
@@ -6,6 +6,7 @@ defmodule PaEss.Utilities do
   require Logger
 
   @space "21000"
+  @stopped_regex ~r/Stopped (\d+) stops? away/
 
   @abbreviation_replacements [
     {~r"\bOL\b", "Orange Line"},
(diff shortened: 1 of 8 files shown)


lib/pa_ess/utilities.ex
```diff
@@ -509,6 +510,46 @@ defmodule PaEss.Utilities do
...
+      min > 60 -> {60, true}
+      prediction.type == :reverse and min > 20 -> {div(min, 10) * 10, true}
+      true -> {max(min, 1), false}
+    end
+  end
+
+  @spec prediction_approaching?(Predictions.Prediction.t(), boolean()) :: boolean()
+  def prediction_approaching?(prediction, terminal?) do
+    !terminal? and !prediction.stopped_at_predicted_stop? and
+      prediction_seconds(prediction, terminal?) in 31..60
```
PaulJKim, line 538: Technically `terminal?` can only ever be `false` when `predictions_seconds/2` is called here. Should we just explicitly pass `false` as the second param?

panentheos: I kind of went back and forth on that. Technically we could just hardcode it, but it creates a little bit of unnecessary coupling in the logic, so I'm inclined to leave it this way.

PaulJKim (approved):
Looks good!

PR #868, opened 2025-01-07
[bugfix] Account for 3-part node descriptions

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Process ARINC JSON for LivePA connectivity into Splunk](https://app.asana.com/0/1185117109217413/1208851406631730/f)

After deploying the initial change, we lost logging of some device types. I think this is due to the description screen for these devices only having three parts to them rather than 4. Oddly, I'm not able to find any `MatchError` logs in Splunk which is what I would expect, but the `Endpoint` process is crashing every 2 minutes which is the cadence at which we receive device status logs.

#### Reviewer Checklist

### DIFF
--- a/lib/monitoring/uptime.ex
+++ b/lib/monitoring/uptime.ex
@@ -77,7 +77,8 @@ defmodule Monitoring.Uptime do
   defp get_log_fields(%{"description" => description, "is_online" => is_online} = node) do
     node_type = get_node_type(node)
 
-    [line, station, device_id, maybe_sign_zone] = String.split(description, ":")
+    [line, station, device_id, maybe_sign_zone] =
+      String.split(description, ":") |> Enum.concat([nil]) |> Enum.take(4)
 
     log_fields = [
       tag: "device_uptime",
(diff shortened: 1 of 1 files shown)


lib/monitoring/uptime.ex
```diff
@@ -77,7 +77,8 @@ defmodule Monitoring.Uptime do
   defp get_log_fields(%{"description" => description, "is_online" => is_online} = node) do
     node_type = get_node_type(node)
 
-    [line, station, device_id, maybe_sign_zone] = String.split(description, ":")
+    [line, station, device_id, maybe_sign_zone] =
+      String.split(description, ":") |> Enum.concat([nil]) |> Enum.take(4)
```
digitalcora, line 81: It's not super clear what the intent is here from reading it — maybe something like this?

```elixir
[line, station, device_id | rest] = String.split(description, ":")
maybe_sign_zone = List.first(rest)
```

`List.first` defaults to returning `nil` when the list is empty.

PaulJKim: works for me! definitely reads easier

digitalcora (approved):

PR #869, opened 2025-01-13
aggregate PA message playback

panentheos:
#### Summary of changes

**Asana Ticket:** [Revise RTS to remove PA Message bottleneck failures in Legacy PA/ESS](https://app.asana.com/0/1185117109217413/1208950993234729/f)

This changes the playback code for PA messages to send a single aggregate message to all targeted zones within a station, instead of individual per-sign messages. This involves moving the responsibility for playback into the PA message engine, and having the signs just provide their configs (and continue to track playbacks in order to guard against accidental duplicates).

### DIFF
--- a/lib/engine/pa_messages.ex
+++ b/lib/engine/pa_messages.ex
@@ -28,9 +28,9 @@ defmodule Engine.PaMessages do
     state =
       case get_active_pa_messages() do
         {:ok, pa_messages} ->
-          recent_sends = send_pa_messages(pa_messages, state.pa_messages_last_sent)
-
-          %{state | pa_messages_last_sent: recent_sends}
+          {selected_pa_messages, state} = select_pa_messages(pa_messages, state)
+          play_pa_messages(selected_pa_messages)
+          state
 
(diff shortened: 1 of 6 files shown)


lib/engine/pa_messages.ex
```diff
@@ -48,47 +48,39 @@ defmodule Engine.PaMessages do
...
-        String.to_existing_atom("Signs/#{sign_id}"),
-        {:play_pa_message, pa_message}
-      )
+  defp play_pa_messages(pa_messages) do
+    Enum.flat_map(pa_messages, fn message ->
+      for sign_id <- message.sign_ids,
+          {sign, should_play?} = GenServer.call(:"Signs/#{sign_id}", {:play_pa_message, message}),
+          should_play? do
+        sign
+      end
```
PaulJKim, line 71: Any thoughts on calling out to all the sign processes concurrently using `Task`?

panentheos: I think using `Task` would be overkill here, since this call should complete more or less instantly.

PaulJKim (approved):
This looks good to me!

PR #870, opened 2025-01-13
unify prediction audios

panentheos:
#### Summary of changes

This combines the `NextTrainCountdown`, `FollowingTrain`, and `StoppedTrain` audio modules into one unified module. The goal is to reduce some overlapping logic between these code paths, and pave the way for future refactoring.

This includes a slight wording change on JFK platforms readouts, which was ok'd by product and is illustrated with a new test case. The other test deltas are minor capitalization and punctuation changes to TTS strings resulting from standardizing the code, which shouldn't substantially affect the output.


lib/content/audio/predictions.ex
```diff
@@ -0,0 +1,137 @@
...
+    defp platform_status(audio) do
+      {minutes, _} = PaEss.Utilities.prediction_minutes(audio.prediction, audio.terminal?)
+      platform = Content.Utilities.stop_platform(audio.prediction.stop_id)
+      jfk_mezzanine? = audio.special_sign == :jfk_mezzanine
+
+      cond do
+        !platform -> {nil, false, nil}
+        jfk_mezzanine? and minutes > 9 -> {nil, false, :later}
+        jfk_mezzanine? and minutes > 5 -> {nil, false, :soon}
+        minutes == 1 or !jfk_mezzanine? -> {platform, true, nil}
```
PaulJKim, line 129: Is the `!jfk_mezzanine?` check needed here? If the prediction is not for either of the JFK NB platforms, then `platform` will be `nil` and the first clause of the `cond` will be returned.

lib/content/audio/predictions.ex
```diff
@@ -0,0 +1,137 @@
...
+            {platform :: Content.platform() | nil, platform_prefix? :: boolean(),
+             platform_tbd :: :later | :soon | nil}
+    defp platform_status(audio) do
+      {minutes, _} = PaEss.Utilities.prediction_minutes(audio.prediction, audio.terminal?)
+      platform = Content.Utilities.stop_platform(audio.prediction.stop_id)
+      jfk_mezzanine? = audio.special_sign == :jfk_mezzanine
+
+      cond do
+        !platform -> {nil, false, nil}
+        jfk_mezzanine? and minutes > 9 -> {nil, false, :later}
```
PaulJKim, line 127: With documentation/posterity (although I think the typespec mostly covers that) and avoiding "magic numbers" in code in mind, it might be worthwhile if we swap out `9` and `5` for some module attributes. `@announce_platform_later_threshold` and `@announce_platform_soon_threshold` or something along those lines?

lib/content/audio/predictions.ex
panentheos: We can get here either by being on the mezzanine with a low `minutes` value, or by being on one of the NB platforms. We want the special `platform_prefix?` behavior to always trigger on the platform (new behavior), and trigger on the mezzanine only at 1 minute (this was existing behavior).

PaulJKim: ah that's right, forgot we had discussed that. Makes sense!

PaulJKim (approved):
Looks good!

PR #873, opened 2025-01-16
introduce Message structs

panentheos:
#### Summary of changes

This introduces a new `Message` protocol and associated structs as part of the ongoing refactoring effort. These new structs represent a piece of sign content at a purely semantic level, in contrast with the existing `Content.Message` structs which encode more concrete, line-level content. The new protocol allows these abstract messages to be "rendered" into various concrete visual (and eventually audio) forms.

Conceptually, message generation is broken into several steps: First, generate the abstract messages to display. Then, optionally transform these messages, e.g. to combine headway or alert statuses. Finally, render the messages into actual line content, including flipping, shortening, or paging as needed.

One immediate benefit of this change is that the message generation algorithm is no longer split between the platform and mezzanine cases. A followup PR will leverage this new framework to generate audio messages as well.

Note that although we still assume we're dealing with at most 2 messages, nothing about this approach specifically requires that. It would be reasonably straightforward to support more messages at some point by rewriting `transform_messages` and `render_messages` accordingly.

### DIFF
--- a/lib/message.ex
+++ b/lib/message.ex
@@ -1,0 +1,13 @@
+defprotocol Message do
+  @doc "render a single line of content, may include line-level paging"
+  @spec to_single_line(Message.t()) :: Content.Message.t()
+  def to_single_line(message)
+
+  @doc "render two lines of content, must not include line-level paging"
+  @spec to_full_page(Message.t()) :: {Content.Message.t(), Content.Message.t()}
+  def to_full_page(message)
+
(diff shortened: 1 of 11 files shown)


PaulJKim:
This is looking great to me! I'll take another look in the AM during the usual code review block but not seeing any concerns

PaulJKim (approved):
🧹 Looks great!

PR #874, opened 2025-01-31
wire up message priority

panentheos:
#### Summary of changes

**Asana Ticket:** [Fix announcement priority for migrated SCUs](https://app.asana.com/0/1185117109217422/1209272297071108/f)

This passes the required priority value to Scully when playing messages. Priority `2` is default, except PA messages which specify their own priority. Also fixes incorrect usage of `https`, which should be `http`

### DIFF
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -56,6 +56,7 @@ defmodule PaEss.Updater do
               Signs.Realtime.t() | Signs.Bus.t(),
               [Content.Audio.value()],
               [Content.Audio.tts_value()],
+              integer(),
               [keyword()]
             ) ::
               :ok
@@ -68,6 +69,7 @@ defmodule PaEss.Updater do
         },
         audios,
         tts_audios,
+        priority,
         log_metas
       ) do
     tags = Enum.map(audios, fn _ -> create_tag() end)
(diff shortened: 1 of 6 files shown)


lib/pa_ess/updater.ex
```diff
@@ -110,6 +112,7 @@ defmodule PaEss.Updater do
                audio_zones: audio_zones,
                audio_data: [Base.encode64(file)],
                expiration: 30,
+               priority: priority,
```
digitalcora, line 115: Would it make sense for the message format here to be a struct instead of a map? That would prevent (/would have prevented?) things like not including a `priority` field when we expect one.

lib/signs/utilities/audio.ex
```diff
@@ -365,6 +365,10 @@ defmodule Signs.Utilities.Audio do
       sign,
       Enum.map(audios, &Content.Audio.to_params(&1)),
       Enum.map(audios, &Content.Audio.to_tts(&1)),
+      case audios do
+        [%PaMessages.PaMessage{priority: priority}] -> priority
+        _ -> 2
```
digitalcora, line 370: In what situations do we have multiple `audios`? Is it okay that we're then ignoring their individual priorities?

panentheos: For PA messages, there's only ever one at a time, so this should be sufficient. A handful of other code paths can generate multiple.

lib/pa_ess/updater.ex
digitalcora: Mm, if this is ultimately the data we're sending directly out of the app to Scully, then it's probably not that important for it to be a struct. I wasn't sure if we did anything with it within RTS, before that.

PR #875, opened 2025-01-31
use message structs to compute audio

panentheos:
#### Summary of changes

This is part of the ongoing message refactoring effort. It uses the recently added `Message` structs to compute audio readouts and announcements. This allows us to remove a couple dense code paths that deal with "decoding" sign content, since we can now use the higher-level Message structs directly. There should be no user-facing changes.


PaulJKim (commented):
One nit. Otherwise looks great!

lib/signs/utilities/audio.ex
```diff
@@ -137,70 +50,71 @@ defmodule Signs.Utilities.Audio do
...
-                (announce_boarding?(sign, message) ||
-                   (announce_arriving?(sign, message) &&
-                      message.prediction.trip_id not in sign.announced_arrivals)) ->
-              {Audio.TrainIsBoarding.from_message(message),
-               update_in(sign.announced_boardings, &cache_value(&1, message.prediction.trip_id))}
+            minutes == :boarding &&
+              prediction.trip_id not in sign.announced_boardings &&
+                (announce_boarding?(sign, prediction) ||
+                   (announce_arriving?(sign, prediction) &&
+                      prediction.trip_id not in sign.announced_arrivals)) ->
```
PaulJKim, line 68: Would it be worth taking the opportunity to separate out some of these long conditions into helpers for readability's sake?

panentheos: That's a good idea. Or something like pulling out each announcement clause, since they're basically independent. I'll experiment with that a bit in a separate PR.

PaulJKim (approved):
Tested locally. Great stuff! 🚢

PR #877, opened 2025-02-06
add carriage details to prediction

panentheos:
#### Summary of changes

**Asana Ticket:** [Special message for 4car RL consists](https://app.asana.com/0/1185117109217422/1209181549811302)

This attaches the `multi_carriage_details` field from the location struct to the prediction, so we can use it for checking for 4-car trains and new Red Line cars. This is straightforward because we already look up the location data when generating each prediction.


digitalcora (approved):
🚃🚃🚃🚃(🚃🚃)

PR #880, opened 2025-02-07
streamline approaching announcement

panentheos:
#### Summary of changes

**Asana Ticket:** [Change timing/existence of approaching/arriving messages](https://app.asana.com/0/1185117109217413/1209206869122456)

This is a preliminary change in support of both the approaching/arriving consolidation and the new four-cars announcements. It unifies the code paths for creating approaching audio, and removes some extraneous checks. Note that while this removes some RL-specific guards in this code, the upstream code already does a similar restriction, so the end result is the same.

The slight change in output is the result of adding an extra space to the message for consistency, which shouldn't have any visible impact after the message is formatted.


digitalcora (approved):
🧹

PR #881, opened 2025-02-10
use new audio takes for approaching message

panentheos:
#### Summary of changes

**Asana Ticket:** [Switch to Polly approaching messages and add yellow line message](https://app.asana.com/0/1185117109217413/1209206869122457)

This switches to the new polly-generated audio takes for approaching messages. Includes the following behavior changes:
* Uses the new standard "[destination] train" phrasing
* Includes both platform and "new cars" modifiers if applicable
* The migrated code no longer does custom formatting, for a closer match with the ARINC behavior (we can change this later).

### DIFF
--- a/lib/content/audio/approaching.ex
+++ b/lib/content/audio/approaching.ex
@@ -33,8 +33,14 @@ defmodule Content.Audio.Approaching do
   end
 
   defimpl Content.Audio do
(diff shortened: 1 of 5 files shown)


digitalcora (approved):
🔉👂✅

PR #882, opened 2025-02-11
remove arriving message and move approaching earlier

panentheos:
#### Summary of changes

**Asana Ticket:** [Change timing/existence of approaching/arriving messages](https://app.asana.com/0/1185117109217413/1209206869122456)

This removes the "arriving" message, changes the "approaching" threshold from 60 to 45 seconds, and enables approaching messages for light rail.


lib/signs/realtime.ex
```diff
@@ -30,9 +30,7 @@ defmodule Signs.Realtime do
               [
                 :headway_stop_id,
                 :current_time_fn,
-                announced_arrivals: [],
                 announced_approachings: [],
-                announced_approachings_with_crowding: [],
```
digitalcora, line 35: What was this used for?

lib/signs/utilities/audio.ex
```diff
@@ -65,27 +64,14 @@ defmodule Signs.Utilities.Audio do
               prediction.trip_id not in sign.announced_boardings &&
                 (announce_boarding?(sign, prediction) ||
                    (announce_arriving?(sign, prediction) &&
```
digitalcora, line 66: Is it still appropriate to call this `announce_arriving?`

lib/signs/utilities/audio.ex
```diff
@@ -179,21 +158,13 @@ defmodule Signs.Utilities.Audio do
...
-          case audio_type do
-            Audio.Approaching -> "approaching"
-            Audio.TrainIsArriving -> "arrival"
-          end
-
+        route_id: "Orange"
+      } ->
         Logger.info(
-          "crowding_log: announcement_type=#{announcement_type} trip_id=#{trip_id} sign_id=#{sign_id} crowding_description=#{inspect(crowding_description)}"
+          "crowding_log: announcement_type=approaching trip_id=#{trip_id} sign_id=#{sign_id} crowding_description=#{inspect(crowding_description)}"
```
digitalcora, line 167: Is `announcement_type` used in some existing report/dashboard or can we get rid of it?

test/content/audio/approaching_test.exs
```diff
@@ -19,14 +19,12 @@ defmodule Content.Audio.ApproachingTest do
                {:canned, {"110", spaced(["896", "915", "920", "910", "21014"]), :audio_visual}}
     end
 
-    test "Returns nil for Green Line trips" do
+    test "Returns params for Green Line trips" do
       audio = %Approaching{destination: :riverside, route_id: "Green-D"}
-      assert Content.Audio.to_params(audio) == nil
-    end
 
-    test "Returns nil when destination is Ashmont on the Mattapan line" do
```
digitalcora, line 27: Do you remember why this special case existed? Or was it not really a special case because we weren't playing Approaching on the Mattapan line at all?

lib/signs/realtime.ex
panentheos: This was to record whether we had included a crowding message in the approaching announcement, and if not, we'd include it in the arriving message. Now that arriving is gone, this isn't relevant (doubly so, since crowding is disabled anyway)

lib/signs/utilities/audio.ex
panentheos: Not entirely, but the config value is still named that, and I didn't want to try and refactor that as part of this change.

lib/signs/utilities/audio.ex
panentheos: I didn't look into it, so this seemed like the safest change. Definitely worth looking at, at some point.

test/content/audio/approaching_test.exs
panentheos: This was part of the general suppression of approaching messages for light rail. I assume the test was to verify the distinction between Ashmont RL and Ashmont Mattapan.

PR #883, opened 2025-02-11
announce four-car trains

panentheos:
#### Summary of changes

**Asana Ticket:** [Approaching announcement for 4-car trains](https://app.asana.com/0/1185117109217413/1209274790976903)
**Asana Ticket:** [New PA/ESS state when next train is 4-car](https://app.asana.com/0/1185117109217413/1209294078164624)

This modifies the approaching announcement and prediction rendering to include special messages for 4-car trains. See tickets for full details.

### DIFF
--- a/lib/content/audio/predictions.ex
+++ b/lib/content/audio/predictions.ex
@@ -1,11 +1,12 @@
 defmodule Content.Audio.Predictions do
-  @enforce_keys [:prediction, :special_sign, :terminal?, :next_or_following]
+  @enforce_keys [:prediction, :special_sign, :terminal?, :multiple?, :next_or_following]
   defstruct @enforce_keys
 
   @type t :: %__MODULE__{
           prediction: Predictions.Prediction.t(),
           special_sign: :jfk_mezzanine | :bowdoin_eastbound | nil,
           terminal?: boolean(),
+          multiple?: boolean(),
           next_or_following: :next | :following
         }
 
@@ -13,14 +14,22 @@ defmodule Content.Audio.Predictions do
           Predictions.Prediction.t(),
           :jfk_mezzanine | :bowdoin_eastbound | nil,
           boolean(),
+          boolean(),
           :next | :following
         ) :: [t()]
-  def new(%Predictions.Prediction{} = prediction, special_sign, terminal?, next_or_following) do
+  def new(
+        %Predictions.Prediction{} = prediction,
+        special_sign,
+        terminal?,
+        multiple?,
+        next_or_following
+      ) do
     [
       %__MODULE__{
         prediction: prediction,
         special_sign: special_sign,
         terminal?: terminal?,
+        multiple?: multiple?,
         next_or_following: next_or_following
       }
     ]
@@ -74,8 +83,9 @@ defmodule Content.Audio.Predictions do
           end
 
         {prefix, suffix} = if(platform_prefix?, do: {qualifier, []}, else: {[], qualifier})
+        four_cars = if four_cars?(audio), do: [:four_car_train_message], else: []
 
-        [the_next_or_following] ++ train ++ prefix ++ status ++ suffix ++ followup
+        [the_next_or_following] ++ train ++ prefix ++ status ++ suffix ++ followup ++ four_cars
       end
       |> PaEss.Utilities.audio_message()
     end
@@ -120,7 +130,9 @@ defmodule Content.Audio.Predictions do
               _ -> ""
             end
 
-          "The #{next_or_following} #{train}#{prefix} #{status}#{suffix}.#{followup}"
+          four_cars = if four_cars?(audio), do: PaEss.Utilities.four_cars_text(), else: ""
+
+          "The #{next_or_following} #{train}#{prefix} #{status}#{suffix}.#{followup}#{four_cars}"
         end
 
       {text, nil}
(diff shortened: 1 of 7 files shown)


lib/content/audio/predictions.ex
```diff
@@ -74,8 +83,9 @@ defmodule Content.Audio.Predictions do
           end
 
         {prefix, suffix} = if(platform_prefix?, do: {qualifier, []}, else: {[], qualifier})
+        four_cars = if four_cars?(audio), do: [:four_car_train_message], else: []
 
-        [the_next_or_following] ++ train ++ prefix ++ status ++ suffix ++ followup
+        [the_next_or_following] ++ train ++ prefix ++ status ++ suffix ++ followup ++ four_cars
```
digitalcora, line 88: `followup` here is possibly something like "we will announce the platform for boarding soon", so does it make sense to then say "move to the front of the platform to board", when riders don't yet know _which_ platform they should move to the front of?

lib/message/predictions.ex
```diff
@@ -37,13 +46,18 @@ defmodule Message.Predictions do
         |> Enum.uniq()
         |> length() == 1
 
-      Enum.take(message.predictions, if(multiple?, do: 1, else: 2))
+      four_cars? =
+        hd(message.predictions) |> PaEss.Utilities.prediction_four_cars?() and !multiple? and
+          !message.terminal?
```
digitalcora, line 51: We're doing this same `four_cars? and !multiple? and !terminal?` conditional in two different places — I think it's actually a different struct in each, so maybe there's no easy way to avoid this, unless they already implement a common protocol/behaviour.

lib/pa_ess/utilities.ex
```diff
@@ -563,6 +567,11 @@ defmodule PaEss.Utilities do
     String.to_integer(str)
   end
 
+  @spec prediction_four_cars?(Predictions.Prediction.t()) :: boolean()
+  def prediction_four_cars?(prediction) do
+    prediction.route_id == "Red" and match?([_, _, _, _], prediction.multi_carriage_details)
```
digitalcora, line 572: nit: I think I like the way we do this in `Message.Predictions.to_multi_line/1` better.

```elixir
def prediction_four_cars?(%Prediction{route_id: "Red", multi_carriage_details: [_, _, _, _]}), do: true
def prediction_four_cars?(_prediction), do: false
```

lib/content/audio/predictions.ex
```diff
@@ -1,26 +1,35 @@
 defmodule Content.Audio.Predictions do
-  @enforce_keys [:prediction, :special_sign, :terminal?, :next_or_following]
+  @enforce_keys [:prediction, :special_sign, :terminal?, :multiple?, :next_or_following]
   defstruct @enforce_keys
 
   @type t :: %__MODULE__{
           prediction: Predictions.Prediction.t(),
           special_sign: :jfk_mezzanine | :bowdoin_eastbound | nil,
           terminal?: boolean(),
+          multiple?: boolean(),
```
digitalcora, line 9: What does `multiple?` mean in this context? I'm not quite able to figure out when this would ultimately be set and when it wouldn't, and why we care about that. Are there test cases that cover both possibilities?

lib/content/audio/predictions.ex
panentheos: Fortunately (?) that followup message will only trigger on the JFK mezzanine, where we don't do passive readouts for four-car predictions. So as of now, this won't be an issue, but this is a good callout because there's a lot of stuff piling up here, and we could easily back ourselves into some nonsensical messaging if we're not careful about it.

lib/message/predictions.ex
panentheos: Yeah, I thought about that, but the context is a bit different in each place, so there wasn't an obvious way to abstract it without making things less clear.

lib/content/audio/predictions.ex
panentheos: It's whether there are multiple messages on the sign, which in this case determines whether we show (and read) the four-car message. Perhaps `multiple_messages?` would be better. Also, there isn't currently a negative test for that case. I can add one.

digitalcora: > It's whether there are multiple messages on the sign, which in this case determines whether we show (and read) the four-car message.

I'm confused by this — don't we always show the 4-car message if the next train is 4 cars, and it overrides a second prediction message if there is one?

lib/content/audio/predictions.ex
digitalcora: > that followup message will only trigger on the JFK mezzanine, where we don't do passive readouts for four-car predictions

Since I didn't see JFK-specific logic in these changes, I'm guessing this is because of a general property of mezzanine signs, but I'm not quite connecting the dots.

lib/content/audio/predictions.ex
panentheos: Sorry, I didn't explain that properly. There's a (newish) concept of a `Message`, which is generated per-config, so platform signs produce one, and mezzanine signs produce two (essentially). A single `Message.Predictions` can then render up to two actual prediction strings (or a single prediction with a four-cars modifier, now). A pair of `Message.Predictions` will render one prediction each, and will skip the four-cars logic.

digitalcora: Ah, I think I see now, and that also answers my question about the JFK mezzanine.

digitalcora: I'd agree that maybe `multiple_messages?` would be a better name — in the absence of a noun for "multiple" to refer to, I'd assume it was "this" module/struct (the prediction audio), but it sounds like it's not that.

digitalcora (approved):
All aboard the merge train 🚃🚃🚃🚃 (it has 4 cars. please move to the front)

I like how this turned out pretty straightforward, with the prior refactors in place!

PR #887, opened 2025-02-25
remove Content.Message

panentheos:
#### Summary of changes

This is part of the ongoing message refactor. Now that audio messages are calculated directly from the new `Message` structs, the only thing we're doing with `Content.Message` structs is immediately calling `to_string` on them. This removes that indirection, and inlines the content-generation code directly in the `Message`. This ends up removing a lot of wiring code, and makes it easier to understand what gets output for each message.

Note that this removes a fair bit of low-level test code for these structs. Most of the cases are already covered by the suite of high-level behavior tests, and I added a few more to cover the ones that weren't.


test/signs/realtime_test.exs
```diff
@@ -1584,6 +1636,27 @@ defmodule Signs.RealtimeTest do
       Signs.Realtime.handle_info(:run_loop, @multi_route_mezzanine_sign)
     end
 
+    test "mezzzanine sign, shuttle alert flips to bottom" do
```
PaulJKim, line 1639: nit: slight spelling error on 'mezzanine' here

PaulJKim (approved):
Tested locally and looks good!

PR #891, opened 2025-02-27
add platform edge message to approaching announcement

panentheos:
#### Summary of changes

**Asana Ticket:** [Add "platform edge" message](https://app.asana.com/0/1185117109217413/1209206869122457)

This adds a "Please stand back" message onto the approaching announcement.

### DIFF
--- a/lib/content/audio/approaching.ex
+++ b/lib/content/audio/approaching.ex
@@ -54,6 +54,8 @@ defmodule Content.Audio.Approaching do
     @with_all_new_red_line_cars "893"
     # audio: "It is a shorter 4-car train. Move toward the front of the train to board, and stand back from the platform edge.", visual: "Please move to front of the train to board."
     @four_car_train_message "922"
+    # "Please stand back from the platform edge."
+    @stand_back_message "925"
     @comma "21012"
     @period "21014"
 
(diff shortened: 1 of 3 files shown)


lib/content/audio/approaching.ex
```diff
@@ -54,6 +54,8 @@ defmodule Content.Audio.Approaching do
     @with_all_new_red_line_cars "893"
     # audio: "It is a shorter 4-car train. Move toward the front of the train to board, and stand back from the platform edge.", visual: "Please move to front of the train to board."
     @four_car_train_message "922"
+    # "Please stand back from the platform edge."
+    @stand_back_message "925"
```
digitalcora, line 58: I take it (heh) we sent this to ARINC as a new take, rather than telling them to overwrite the take we previously expected to use, 898 "please stand back from the yellow line"?

test/content/audio/approaching_test.exs
```diff
@@ -9,30 +9,33 @@ defmodule Content.Audio.ApproachingTest do
 
       assert Content.Audio.to_params(audio) ==
                {:canned,
-                {"112", spaced(["896", "892", "920", "910", "901", "21014"]), :audio_visual}}
+                {"114", spaced(["896", "892", "920", "910", "901", "21014", "925"]),
+                 :audio_visual}}
```
digitalcora, line 13: Did we not already have a test case in this file for the "four cars" message? I don't see any reference to that particular take, 922.

digitalcora (approved):
LGTM, I had a couple of non-blocking questions.

test/content/audio/approaching_test.exs
panentheos: Not in this file. Generally we've been preferring the higher level behavior tests because they give a lot more coverage and resilience for about the same amount of code. `realtime_test.exs` has the cases for that message.

PR #892, opened 2025-02-27
remove outdated dialyzer config

panentheos:
#### Summary of changes

This cleans up some unnecessary/incorrect dialyzer options. This will run dialyzer with a default configuration, which should be suitable.


digitalcora (approved):
🧹

PR #894, opened 2025-02-28
remove Sentry

panentheos:
#### Summary of changes

We currently install and configure Sentry, but we're not actually using it for this project, and all of our monitoring is handled elsewhere (e.g. Splunk). Perhaps we should be using it to capture errors, but since it's not currently doing anything, let's remove it to create a clean slate, with the intention of wiring it up properly later.


digitalcora (approved):
🪓

PR #896, opened 2025-03-04
upgrade github action versions

panentheos:
#### Summary of changes

The old version of the `cache` action is deprecated and failing.


digitalcora (approved):
There are also some newer versions of our other action dependencies we could be using, see e.g. mbta/screenplay#610.

panentheos:
Yep. Also this is an older code base so I suspect there's some additional cleanup of the action code we could do.

PR #897, opened 2025-03-07
refactor take messages

panentheos:
#### Summary of changes

This fully standardizes the handling of audio take ids. All take ids are now produced by the `audio_take` function, which maps a semantic-but-readable identifier to the corresponding numeric take id. The higher-level `audio_message` can now optionally produce `:audio_visual` messages, allowing it to take over the remaining cases that were not already using this pattern.

### DIFF
--- a/lib/pa_ess/utilities.ex
+++ b/lib/pa_ess/utilities.ex
@@ -160,13 +160,6 @@ defmodule PaEss.Utilities do
     Integer.to_string(5000 + n)
   end
 
-  @doc "Constructs message from TAKE variables"
-  @spec take_message([String.t()], Content.Audio.av_type()) :: Content.Audio.canned_message()
-  def take_message(vars, av_type) do
-    vars_with_spaces = pad_takes(vars)
-    {:canned, {take_message_id(vars_with_spaces), vars_with_spaces, av_type}}
-  end
-
   @doc "Intersperse spaces in a list of takes, accounting for punctuation"
   @spec pad_takes([String.t()]) :: [String.t()]
   def pad_takes(vars) do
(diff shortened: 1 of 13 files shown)


lib/pa_ess/utilities.ex
```diff
@@ -782,18 +740,104 @@ defmodule PaEss.Utilities do
...
+  def audio_take({:line, "Green"}), do: "3008"
+  def audio_take({:line, "Mattapan"}), do: "3009"
+  def audio_take({:line, _}), do: audio_take(:train)
+
+  def audio_take({:passthrough, :alewife, _}), do: "32114"
+  def audio_take({:passthrough, :southbound, "Red"}), do: "891"
+  def audio_take({:passthrough, :bowdoin, _}), do: "32111"
+  def audio_take({:passthrough, :wonderland, _}), do: "32110"
+  def audio_take({:passthrough, :forest_hills, _}), do: "32113"
+  def audio_take({:passthrough, :oak_grove, _}), do: "32112"
```
digitalcora, line 833: Can we name the ignored parts of the pattern so it's clear what it is? This looks like a `_route_id` for example. (the `:crowding` ones are especially unclear to me)

lib/pa_ess/utilities.ex
```diff
@@ -782,18 +740,104 @@ defmodule PaEss.Utilities do
...
+
+  def audio_take(_), do: nil
```
digitalcora, line 835: Is it important to have this fallback? Or rather, what does `nil` end up doing here? If it "fails safe", then we'll never know we have a logic error since this also doesn't log a warning. If it will ultimately cause a crash, then it seems better to omit the fallback and let the crash happen early to make it clear what the problem is.

lib/pa_ess/utilities.ex
```diff
@@ -661,110 +605,124 @@ defmodule PaEss.Utilities do
...
-    bus_service_to: "877",
-    no_bus_service: "878",
-    b: "536",
-    c: "537",
-    d: "538",
-    e: "539"
-  }
+  @spec audio_take(term()) :: String.t() | nil
+  def audio_take(:alewife), do: "4000"
+  def audio_take(:alewife_), do: "892"
```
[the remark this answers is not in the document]
panentheos: Underscore versions are the newer Polly-generated ones. Perhaps a comment at the top would be helpful?

lib/pa_ess/utilities.ex
digitalcora: Ah, thanks for that context that this does generate a warning. What do you think of putting the warning here, where the "failure" happens? That way it would be impossible to use this function improperly _without_ logging a warning.

lib/pa_ess/utilities.ex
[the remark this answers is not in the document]
digitalcora: Hm, I see, though that is a somewhat obtuse naming convention. I'd say it at least merits a comment, yeah.

lib/pa_ess/utilities.ex
```diff
@@ -661,110 +605,125 @@ defmodule PaEss.Utilities do
...
-    no_service: "879",
-    there_is_no: "880",
-    bus_service_to: "877",
-    no_bus_service: "878",
-    b: "536",
-    c: "537",
-    d: "538",
-    e: "539"
-  }
+  @spec audio_take(term()) :: String.t() | nil
```
digitalcora, line 608: With the latest change, this now never returns `nil`.

digitalcora (approved):
Nice! ✨

PR #898, opened 2025-03-13
Don't announce 4 cars when following train

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Only announce 4-car trains for next trains, not following trains](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1209554302146849)

Only include the 4-car messaging when the announcement is for the next train, not the following train.

#### Reviewer Checklist

### DIFF
--- a/lib/content/audio/predictions.ex
+++ b/lib/content/audio/predictions.ex
@@ -1,5 +1,12 @@
 defmodule Content.Audio.Predictions do
-  @enforce_keys [:prediction, :special_sign, :terminal?, :multiple_messages?, :next_or_following]
+  @enforce_keys [
+    :prediction,
+    :special_sign,
+    :terminal?,
+    :multiple_messages?,
+    :next_or_following,
+    :is_first?
+  ]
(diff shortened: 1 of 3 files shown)


lib/content/audio/predictions.ex
```diff
@@ -161,7 +161,7 @@ defmodule Content.Audio.Predictions do
 
     defp four_cars?(audio) do
       PaEss.Utilities.prediction_four_cars?(audio.prediction) and !audio.terminal? and
-        !audio.multiple_messages?
+        !audio.multiple_messages? and audio.next_or_following == :next
```
panentheos, line 164: This would still generate a readout if the second prediction was read with "the next", which would happen if the destinations were different (e.g. Ashmont and Braintree). I think we'll need another flag or something, which we could set in `Message.Predictions.to_audio` based on the index.

PaulJKim: Ah thanks for catching that!

PR #900, opened 2025-03-14
use Logger.warning instead of Logger.warn

panentheos:
#### Summary of changes

This switches to `Logger.warning` instead of the deprecated `Logger.warn`


digitalcora (approved):
🧹

PR #902, opened 2025-03-18
Update ARCHITECURE.md to reflect current state

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Update RTS/SignsUI architecture docs](https://app.asana.com/0/1185117109217413/1209020825682295/f)

Update the RTS architecture doc.

I purposely left out a few pieces that I thought were subject to change in the near-ish future with upcoming migrations. For example, I avoided going into much detail about the message queue since as SCUs are migrated, the existing message queue and corresponding "updater" class will slowly be phased out.

I also removed some details that don't seem as relevant these days as RTS has stabilized over the last couple years. For example, mention about potentially DOSing the ARINC headend server.


panentheos (approved):
Looks good!

PR #906, opened 2025-04-08
Use minimum alert status

PaulJKim:
#### Summary of changes

**Asana Ticket:** [All destinations must be in alert mode to trigger alert status](https://app.asana.com/0/1185117109217413/1208776052650234/f)

This change makes it so that instead of using a sign's "maximum alert status", we use the "minimum alert status". This effectively makes it so that we prefer to show less information rather than more in cases where there are > 1 alert statuses for a given sign.

A couple nuances to the change:
- We should be able to remove the consideration of `overall_route_status` when calculating the minimum alert status. Based on the logic in `lib/engine/alerts/api_fetcher.ex`, routes only receive the status `:alert_along_route` and that alert status isn't useful to us.
- I reordered the alert status priorities based on the principal that more information = higher priority level
- Not a change, but note that alert statuses on a single stop_id granularity will still use a "maximum alert status" since at a single stop level, there is no ambiguity on what information should be displayed.

#### Reviewer Checklist

### DIFF
--- a/lib/engine/alerts.ex
+++ b/lib/engine/alerts.ex
@@ -24,30 +24,16 @@ defmodule Engine.Alerts do
     GenServer.start_link(__MODULE__, opts, name: name)
   end
 
-  @callback max_stop_status([Fetcher.stop_id()], [Fetcher.route_id()]) :: Fetcher.stop_status()
-  def max_stop_status(
+  @callback min_stop_status([Fetcher.stop_id()]) :: Fetcher.stop_status()
+  def min_stop_status(
         tables \\ %{stops_table: @stops_table, routes_table: @routes_table},
-        stop_ids,
-        route_ids
+        stop_ids
       ) do
-    overall_stop_status =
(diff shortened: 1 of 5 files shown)


lib/engine/alerts.ex
```diff
@@ -24,30 +24,16 @@ defmodule Engine.Alerts do
...
-      end)
-
-    if Enum.all?(route_states, fn s -> s == overall_route_status end) do
-      Fetcher.higher_priority_status(overall_stop_status, overall_route_status)
-    else
-      overall_stop_status
-    end
+    stop_ids
+    |> Enum.map(&stop_status(tables.stops_table, &1))
+    |> Enum.reduce(fn current_status, overall_status ->
```
panentheos, line 34: nit: We could use `Enum.min_by` to make this slightly clearer and more concise.

panentheos (approved):
Looks good!

PR #911, opened 2025-04-15
switch to PCM audio generation

panentheos:
#### Summary of changes

**Asana Ticket:** [SCU audio and visual fixes](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209953188678181?focus=true)

This changes the audio generation from `mp3` to `pcm`, to go with the corresponding change in Scully.

### DIFF
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -172,10 +172,14 @@ defmodule PaEss.Updater do
 
     text = ~s(<speak><amazon:effect name="drc">#{text}</amazon:effect></speak>)
 
-    http_poster.post("#{watts_url}/tts", %{text: text, voice_id: voice_id} |> Jason.encode!(), [
-      {"Content-type", "application/json"},
-      {"x-api-key", watts_api_key}
-    ])
+    http_poster.post(
+      "#{watts_url}/tts",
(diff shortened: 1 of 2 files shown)


lib/pa_ess/updater.ex
```diff
@@ -172,10 +172,14 @@ defmodule PaEss.Updater do
 
     text = ~s(<speak><amazon:effect name="drc">#{text}</amazon:effect></speak>)
 
-    http_poster.post("#{watts_url}/tts", %{text: text, voice_id: voice_id} |> Jason.encode!(), [
-      {"Content-type", "application/json"},
-      {"x-api-key", watts_api_key}
-    ])
+    http_poster.post(
+      "#{watts_url}/tts",
+      %{text: text, voice_id: voice_id, output_format: "pcm"} |> Jason.encode!(),
```
digitalcora, line 177: The pipeline here seems superfluous since there's only a single function call, and this might read better:

```elixir
http_poster.post(
  "#{watts_url}/tts",
  Jason.encode!(%{text: text, voice_id: voice_id, output_format: "pcm"}),
  # ...
)
```

digitalcora (approved):
One smol suggestion. Seems good!

PR #912, opened 2025-04-22
[chore] Cutover to new Chelsea Bridge API

robbie-sundstrom:
#### Summary of changes

**Asana Ticket:** [Leverage new Chelsea Bridge API endpoint](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1207065511877770?focus=true)

Hoping to add unit tests for this updated Chelsea Bridge logic.

Adds two API calls
- Call to get a token for the bridge status API, which is then stored in ETS
- Call to the bridge status API with a valid token, which then updates the status of the bridge in ETS

#### Reviewer Checklist


lib/engine/chelsea_bridge.ex
```diff
@@ -22,36 +26,84 @@ defmodule Engine.ChelseaBridge do
...
+           ]),
+         {:ok, data} <- Jason.decode(body) do
+      expiration =
+        now
+        |> DateTime.add(Map.get(data, "expires_in"), :second)
+
+      :ets.insert(
+        :bridge_status,
+        {:auth, %{access_token: Map.get(data, "access_token"), expiration: expiration}}
+      )
```
panentheos, line 101: Note: We use ETS in these engines so that the sign processes can read the data in parallel without creating a bottleneck on the engine process. Since the cached token is only used internally, it would be preferable to store it in state.

robbie-sundstrom: Did modify this token and its expiration to be saved in state!

lib/engine/chelsea_bridge.ex
```diff
@@ -2,56 +2,127 @@ defmodule Engine.ChelseaBridge do
...
 
   def handle_info(:update, state) do
     Process.send_after(self(), :update, 60_000)
-    url = Application.get_env(:realtime_signs, :chelsea_bridge_url)
-    auth = Application.get_env(:realtime_signs, :chelsea_bridge_auth)
     http_client = Application.get_env(:realtime_signs, :http_client)
+    now = Timex.now()
+
+    token =
+      if(state.token.expiration == nil or DateTime.compare(state.token.expiration, now) == :lt) do
```
panentheos, line 60: nit: `Timex` has some slightly more ergonomic functions for comparisons, like `Timex.before?` and `Timex.after?`

lib/engine/chelsea_bridge.ex
```diff
@@ -2,56 +2,127 @@ defmodule Engine.ChelseaBridge do
...
-      :ets.insert(:bridge_status, {:value, %{raised?: status == "Raised", estimate: estimate}})
+           http_client.get("#{@base_api_url}#{@api_status_endpoint}", [
+             {"Authorization", "Bearer #{token.value}"}
+           ]),
+         {:ok, data} <- Jason.decode(body) do
+      EtsUtils.write_ets(
+        state.table,
+        %{
+          :value => %{
+            raised?: Map.get(data, "liftInProgress"),
```
panentheos, line 75: Although it's probably unlikely, these accesses will return `nil` if the data ever looks different from what we expect, which would probably cause a crash. We could add a little extra safety by translating this into a pattern match in the `with` clause above, which would instead let us fall through to the error logging below.

Side nit: In cases where we do just want to get a map value with no default, `data[key]` is equivalent.

lib/engine/chelsea_bridge.ex
```diff
@@ -2,56 +2,127 @@ defmodule Engine.ChelseaBridge do
...
-          _ ->
-            nil
-        end
-
-      :ets.insert(:bridge_status, {:value, %{raised?: status == "Raised", estimate: estimate}})
+           http_client.get("#{@base_api_url}#{@api_status_endpoint}", [
+             {"Authorization", "Bearer #{token.value}"}
+           ]),
+         {:ok, data} <- Jason.decode(body) do
+      EtsUtils.write_ets(
```
panentheos, line 71: nit: `EtsUtils.write_ets` is not strictly necessary here, since it's primarily meant for doing atomic bulk updates of multiple values. A single `:ets.insert` would suffice.

lib/engine/chelsea_bridge.ex
```diff
@@ -2,56 +2,127 @@ defmodule Engine.ChelseaBridge do
...
+            estimate: DateTime.add(now, Map.get(data, "estimatedDurationInMinutes"), :minute)
+          }
+        },
+        :none
+      )
+
+      {:noreply, %{state | token: token}}
     else
       err ->
         Logger.error("Error getting bridge status: #{inspect(err)}")
```
panentheos, line 85: We need a `{:noreply, state}` here since we removed it from the outer code path below.

lib/engine/chelsea_bridge.ex
```diff
@@ -2,56 +2,127 @@ defmodule Engine.ChelseaBridge do
...
+        now
+        |> DateTime.add(Map.get(data, "expires_in"), :second)
+
+      %{
+        value: Map.get(data, "access_token"),
+        expiration: expiration
+      }
+    else
+      err ->
+        Logger.error("Error getting bridge access_token: #{inspect(err)}")
```
panentheos, line 125: We should return `%{value: nil, expiration: nil}` here, to avoid crashing when we try to use the token. It'll still try to make the call with a blank token, and the server should send a failure response, but that seems fine.

lib/engine/chelsea_bridge.ex
robbie-sundstrom: Great point! Updated to pattern match in the second step of the `with` clause as well to directly pull out `liftInProgress` and `estimatedDurationInMinutes`

lib/engine/chelsea_bridge.ex
robbie-sundstrom: Thanks for pointing out! Switched over to use `Timex.before` here, and `Timex.shift` elsewhere, which I now prefer to the DateTime equivalent.

PR #913, opened 2025-04-24
Easier bus config

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Easier configuration for bus PA/ESS](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1208977593203660?focus=true)

This PR allows for an easier way to configure bus signs in Realtime Signs by letting us provide a parent stop id, a list of route ids we care about, and a direction id.

Two core things that are added in this PR:
1. `Engine.BusStops` - some new state that maintains a mapping of parent_stop_id -> relevant child stop ids, based on what route ids and directions we care about. I used the schedules API for this because the stops API alone can only tell us what child stops belong to a parent stop. We need either predictions or schedules to map a parent/route/direction to a child stop id. I attempted building this into the existing `Engine.BusPredictions` GenServer, but since that data is real-time, `stop_ids` could be dropped from the `included` when there are no predictions for a given child stop id which could have effects on our bus alerts logic. Any updates to the Schedules API go hand-in-hand with rating changes so it should line up naturally with times that we need to update bus sign configs due to those.

2. `consolidate_sources` - I added this flag to the config to help with handling the expansion of source configs when the config uses a parent stop id and list of routes. 

Also note: I only updated the configs for the busways called out in the task, but this should technically work for the silver line signs as well if we wanted to change those configs as well.

#### Reviewer Checklist


digitalcora:
> `consolidate_sources` - I added this flag to the config to help with handling the expansion of source configs when the config uses a parent stop id and list of routes.

Could you elaborate on this part? Having read this and taken a quick look at the code, it's still not clear to me what this field actually does, and when we should use it or when we shouldn't.

PaulJKim:
@digitalcora yeah it's confusing and honestly also a bit tough for me to verbalize clearly so apologies in advance if this explanation is hard to follow.

Essentially there are two ways we configure signs to pull predictions from multiple "sources" (a stop_id, route_id, and direction_id combination). We usually either configure a single "source config" object with a list of sources (see the config for bus.Nubian_Platform_A` for example), or we configure a list of "source config" objects, which is more common (see the config for `bus.Harvard_upper`). The prior results in the sign just always showing the two soonest predictions out of the predictions for its list of sources. The latter results in the sign always showing at least one prediction per "source". These two approaches aren't necessarily mutually exclusive, but we sort of use them that way I guess just because of the nature of how we intend each sign to work.

Now going back to the scope of this PR, we always normalize the configs in cases where there's a parent stop id and list of routes provided in the config and we typically want the relevant sources (stop_id, route_id, direction_id) to be considered as it's own "source_config" (the harvard upper busway case). But in some cases, we want them all bundled up as a single "source_config" (nubian platform A case). The aim of this task was to enable a general and simplified way of configuring bus signs using a parent_stop_id and list of route_ids, but we still need a way to distinguish the two types of configurations I outlined above. There doesn't seem to be a way we can infer that difference with this simplified bus config format like we can from the current config shapes, so I felt that adding a simple flag to indicate when we want to bundle up the sources was the most straightforward way to achieve it.

Sorry if that explanation is convoluted. The RTS config is a bit ambiguous in some ways so let me know what else I can try to clarify for you.

digitalcora:
Thanks for that explanation! If I'm understanding it correctly:

1. When we use a parent stop ID, we'd set it up as a single "source config".
2. When `consolidate_sources` is enabled, this results in a single "source config" internally.
3. When `consolidate_sources` is _not_ enabled, this results in _multiple_ "source configs" internally.

I think this would be less confusing if the name and meaning of this flag were the inverse: When it's `false`, one source config results in one source config, sort of as you'd expect by looking at it. When it's `true`, we "expand/split" that single config into multiple. The "enabled" behavior is the one where extra stuff happens, where the resulting config is "more different" from how it looks in the file. Does that make sense?

As for the naming, I'd suggest something like `split_parent_station_sources`, to make it clearer that this only does anything when we're using parent stations, but not sure what the best "verb" is here.

digitalcora:
I think it's less clear what `isolate` would do than what `split` would do. `split` sounds like it takes one "thing" and turns it into multiple things (which is roughly accurate to what's happening here). `isolate` doesn't sound like the number of "things" will change, just that some property of the thing will change. So, between the two, I like `split` better, but I'll agree it's not perfect.

panentheos:
Taking a step back, I'm not sure about the splitting/consolidation logic in general. It does make the configuration more compact, but at the cost of making it less clear at a glance how many "configs" a sign has. It seems to me that: 1. being able to specify the parent stop_id solves the need in the original task, 2. allowing rollups of multiple route_ids _within_ a config seems fine, and will let us collapse a few configs, but 3. programmatically rearranging configs via this new flag feels like a step too far. Curious to hear other thoughts on this!

digitalcora:
From this:

> we typically want the relevant sources (stop_id, route_id, direction_id) to be considered as it's own "source_config" (the harvard upper busway case). But in some cases, we want them all bundled up as a single "source_config" (nubian platform A case). The aim of this task was to enable a general and simplified way of configuring bus signs using a parent_stop_id and list of route_ids, but we still need a way to distinguish the two types of configurations I outlined above.

...I got the impression that this flag was genuinely required, because it's "ambiguous" what to do with a source config that includes a parent station (expand it into multiple sources or multiple source configs) and both have legitimate use cases. Could you outline how we'd handle both the Harvard and Nubian cases @PaulJKim mentioned, if we didn't have this option?

PaulJKim:
@digitalcora well, those are my thoughts as well. I'm not sure how we would handle those two cases without explicitly calling it out in the config since currently the two behaviors are implied by the shape of a sign's config. We can't really do that if we standardize on the same config shape. Alternatively, I suppose we could just tailor the parent stop id case to always split the configs up and only convert the configs for signs that are already using that approach. Then we just leave any signs using grouped configs as is with no parent stop id.

PaulJKim:
@panentheos yeah valid points
> at the cost of making it less clear at a glance how many "configs" a sign has
I wonder how much of a cost this really is though. The original intent of the ticket almost seems like this is sort of the aim - that is, we shouldn't have to be concerned with setting up a number of configs equivalent to how many routes/directions we want displayed on the sign and that we should be able to just give a sign a parent stop, list of routes, a direction, and just let it do its thing.

> programmatically rearranging configs via this new flag feels like a step too far
Yeah I do agree it feels odd to do. To my point in my response to Cora's comment - I'm not sure how we can distinguish the two behaviors without relying on the shape of how a signs sources are organized like we have historically done. Would love to hear any ideas on that front!

An alternative approach I considered that still mostly covers the need of the task is to query and store/key predictions based on parent stop id, if the prediction is for a child of a parent stop that exists in our config. We can keep the configs mostly as is, but at least just swap out the child stop ids. Routes wouldn't be provided in a list in that case and there would still be a separate "source" for each parent_stop_id/route_id/direction_id grouping.

panentheos:
The way I read the ticket, it sounds like the main pain point is having to come up with the corresponding child stop ids for each of the routes. What I was envisioning was keeping the same config structure, and just replacing the child stop id with the parent. That means e.g. Harvard would still have several "configs", but the stop_ids would all be the same parent id. So it's still a bit verbose, but making changes should be similarly straightforward. The differentiation of behavior would work the same as today, based on the number of configs.

The rolling up of several routes _within_ a config is something I hadn't thought about before, but it seems like a nice way to compress the couple of sign configs where that's applicable, without affecting the overall data shape, so that seems like an improvement.

PaulJKim:
@panentheos I see - thanks for the thoughts! That makes sense to me. I'll rework this PR a bit accordingly and ping you when it's ready for another look

PaulJKim:
@panentheos I think this is ready for a pass now. I decided to leave out rolling up the routes into a list because it felt like the slight addition of complexity to handle `route_id` as a string or list wasn't really worth the benefit of us allowing us to compress a couple configs.

lib/engine/bus_stops.ex
```diff
@@ -0,0 +1,105 @@
...
+
+  def handle_info(
+        :update,
+        %{last_modified: last_modified, all_bus_stop_ids: all_bus_stop_ids} = state
+      ) do
+    schedule_update(self())
+    api_url = Application.get_env(:realtime_signs, :api_v3_url)
+    api_key = Application.get_env(:realtime_signs, :api_v3_key)
+    http_client = Application.get_env(:realtime_signs, :http_client)
+    parent_route_directions = SignsConfig.all_bus_stop_route_direction_ids()
```
panentheos, line 39: Since we're doing repeated `in` lookups, we could convert this to a `MapSet` for a performance improvement.

lib/engine/bus_stops.ex
```diff
@@ -0,0 +1,105 @@
...
+           api_url <> "/schedules",
+           Enum.concat(
+             if(last_modified, do: [{"if-modified-since", last_modified}], else: []),
+             if(api_key, do: [{"x-api-key", api_key}], else: [])
+           ),
+           timeout: 10000,
+           recv_timeout: 10000,
+           params: %{
+             "filter[stop]" => Enum.join(all_bus_stop_ids, ","),
+             "filter[route]" => Enum.join(bus_routes, ","),
```
panentheos, line 52: This has a lot of duplicates, so we should `Enum.uniq` it here or above.

panentheos (approved):
See some small comments, then looks good. I like this scoped down version!

lib/engine/bus_stops.ex
PaulJKim: Ah good point. Thank you!

PR #915, opened 2025-04-29
make predictions engine resilient to bad json

panentheos:
#### Summary of changes

`Engine.Predictions` occasionally crashes because the API returns a 200 response with a non-JSON body. This makes the code resilient against that case, so we'll just log a warning instead.


digitalcora (approved):
Curious what the response is in this case, since it evidently didn't fall under the "empty" case that was already handled by `parse_json_response("")`.

panentheos:
I couldn't see the whole response in the log, but the error indicated it starts with `<`, so I'm guessing it's HTML or XML of some sort. In any case, it seems very rare so just glossing over it seems reasonable.

PR #919, opened 2025-05-21
Added logic for overnight period stopping certain messages

deanshi:
#### Summary of changes
**Asana Ticket:** [Suppress some passive content on PA/ESS overnight](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209393352927336?focus=true)
- During this period, alerts and headways will not be displayed or announced
- Predictions, service ended messages and early am messages will still continue to display
- Tests added and modified to adjust to logic, display_headway expected to be called 0 times with the new logic flow

#### Reviewer Checklist

### DIFF
--- a/lib/engine/scheduled_headways.ex
+++ b/lib/engine/scheduled_headways.ex
@@ -85,6 +85,13 @@ defmodule Engine.ScheduledHeadways do
     |> min_time()
   end
 
+  @callback get_last_scheduled_departure([binary]) :: nil | DateTime.t()
+  def get_last_scheduled_departure(stop_ids) do
+    get_first_last_departures(stop_ids)
+    |> Enum.map(&elem(&1, 1))
+    |> min_time()
+  end
+
   @doc "Checks if the given time is after the first scheduled stop and before the last.
   A buffer of minutes (positive) is subtracted from the first time. so that headways are
   shown for a short time before the first train."
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -3,9 +3,11 @@ defmodule Signs.Utilities.Messages do
   Helper functions for deciding which message a sign should
   be displaying
   """
+  alias Signs.Utilities.SourceConfig
 
   @early_am_start ~T[03:29:00]
   @early_am_buffer -40
+  @overnight_buffer -30
 
   @spec get_messages(
           Signs.Realtime.predictions(),
@@ -52,12 +54,19 @@ defmodule Signs.Utilities.Messages do
 
           alert_status = filter_alert_status(alert_status, sign_config)
 
-          prediction_message(predictions, config, sign) ||
-            service_ended_message(service_status, config) ||
-            alert_message(alert_status, sign, config) ||
-            Signs.Utilities.Headways.headway_message(config, current_time) ||
-            early_am_message(current_time, scheduled, config) ||
-            %Message.Empty{}
+          if overnight_period?(current_time, sign.source_config, service_status) do
+            prediction_message(predictions, config, sign) ||
+              service_ended_message(service_status, config) ||
+              early_am_message(current_time, scheduled, config) ||
+              %Message.Empty{}
+          else
+            prediction_message(predictions, config, sign) ||
+              service_ended_message(service_status, config) ||
+              alert_message(alert_status, sign, config) ||
+              Signs.Utilities.Headways.headway_message(config, current_time) ||
+              early_am_message(current_time, scheduled, config) ||
+              %Message.Empty{}
+          end
         end)
     end
     |> transform_messages()
@@ -256,6 +265,42 @@ defmodule Signs.Utilities.Messages do
     if in_early_am?(current_time, scheduled_time) do
       %Message.FirstTrain{destination: config.headway_destination, scheduled: scheduled_time}
     end
+  end
+
+  defp overnight_period?(current_time, config, service_ended) do
+    first_departure =
+      RealtimeSigns.headway_engine().get_first_scheduled_departure(
+        SourceConfig.sign_stop_ids(config)
+      )
+
+    last_scheduled_departure =
+      RealtimeSigns.headway_engine().get_last_scheduled_departure(
+        SourceConfig.sign_stop_ids(config)
+      )
+
+    last_actual_departure =
+      if service_ended do
+        SourceConfig.sign_stop_ids(config)
+        |> Stream.flat_map(&RealtimeSigns.last_trip_engine().get_recent_departures(&1))
+        |> Enum.max_by(fn {_key, datetime} -> datetime end, fn -> nil end)
+        |> case do
+          {_key, dt} -> dt
+          nil -> nil
+        end
+      else
+        nil
+      end
+
+    last_time_to_check =
+      if last_actual_departure != nil &&
+           Timex.after?(last_actual_departure, last_scheduled_departure) do
+        last_actual_departure
+      else
+        last_scheduled_departure
+      end
+
+    Timex.after?(Timex.shift(current_time, minutes: @overnight_buffer), last_time_to_check) ||
+      before_early_am_threshold?(current_time, first_departure)
   end
 
   defp alert_message(alert_status, sign, config) do
--- a/test/signs/realtime_test.exs
+++ b/test/signs/realtime_test.exs
@@ -162,6 +162,10 @@ defmodule Signs.RealtimeTest do
         datetime(~T[05:00:00])
       end)
 
+      stub(Engine.ScheduledHeadways.Mock, :get_last_scheduled_departure, fn _ ->
+        datetime(~D[2023-01-02], ~T[02:00:00])
+      end)
+
       :ok
     end
 
@@ -1227,7 +1231,7 @@ defmodule Signs.RealtimeTest do
       end)
 
       expect_messages({"Southbound train", "due 5:00"})
-      expect(Engine.ScheduledHeadways.Mock, :display_headways?, fn _, _, _ -> false end)
+      expect(Engine.ScheduledHeadways.Mock, :display_headways?, 0, fn _, _, _ -> false end)
 
       Signs.Realtime.handle_info(:run_loop, %{
         @sign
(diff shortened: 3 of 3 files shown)


test/signs/realtime_test.exs
```diff
@@ -162,6 +162,10 @@ defmodule Signs.RealtimeTest do
         datetime(~T[05:00:00])
       end)
 
+      stub(Engine.ScheduledHeadways.Mock, :get_last_scheduled_departure, fn _ ->
+        datetime(~D[2023-01-02], ~T[02:00:00])
```
deanshi, line 166: Just to be clear - the first scheduled departure should _never_ be after the last scheduled departure right? The API will never return that state?

deanshi, line 166: @PaulJKim just wanted to clarify here - we'll never have a case where the first scheduled departure is after the last scheduled departure right? So making this explicit in the tests is okay? 

This comes into play with the `before_early_am_threshold` with the after last scheduled departure logic within `overnight_period?` and wanted to make sure we never get into the weird state where this might happen.

test/signs/realtime_test.exs
```diff
@@ -1305,7 +1309,7 @@ defmodule Signs.RealtimeTest do
       end)
 
       expect(Engine.Predictions.Mock, :for_stop, fn _, _ -> [] end)
-      expect(Engine.ScheduledHeadways.Mock, :display_headways?, 2, fn _, _, _ -> false end)
+      expect(Engine.ScheduledHeadways.Mock, :display_headways?, 0, fn _, _, _ -> false end)
```
deanshi, line 1312: Wanted to just double check here, for early AM and this new logic, we aren't showing headways anymore right?

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +267,42 @@ defmodule Signs.Utilities.Messages do
     end
   end
 
+  defp overnight_period?(current_time, config, service_ended) do
+    first_departure =
+      RealtimeSigns.headway_engine().get_first_scheduled_departure(
```
PaulJKim, line 272: I think we ideally want to try limiting accessing state (calling out to engines) to the main `realtime.ex`. We actually already pass in the first scheduled departure to the `get_messages` function via the `scheduled` variable (admittedly could be named a little better). We can fetch the last scheduled departure as well and pass that along too

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +267,42 @@ defmodule Signs.Utilities.Messages do
...
+        SourceConfig.sign_stop_ids(config)
+      )
+
+    last_scheduled_departure =
+      RealtimeSigns.headway_engine().get_last_scheduled_departure(
+        SourceConfig.sign_stop_ids(config)
+      )
+
+    last_actual_departure =
+      if service_ended do
```
PaulJKim, line 282: `service_ended` could be in the shape of a tuple if it was calculated for a sign that has multiple sources (bidirectional signs) so we'll need to keep that in mind as well

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +267,42 @@ defmodule Signs.Utilities.Messages do
...
+        SourceConfig.sign_stop_ids(config)
+        |> Stream.flat_map(&RealtimeSigns.last_trip_engine().get_recent_departures(&1))
+        |> Enum.max_by(fn {_key, datetime} -> datetime end, fn -> nil end)
+        |> case do
+          {_key, dt} -> dt
+          nil -> nil
+        end
```
PaulJKim, line 289: I think you could simplify this a bit by having the fallback just return {nil, nil} and then piping the result to an `elem(1)`

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +267,42 @@ defmodule Signs.Utilities.Messages do
...
+      else
+        nil
+      end
+
+    last_time_to_check =
+      if last_actual_departure != nil &&
```
PaulJKim, line 295: since `last_actual_departure` is either a datetime or nil, we could do this check as just `if last_actual_departure`

lib/signs/utilities/messages.ex
deanshi: nice 👍

lib/signs/utilities/messages.ex
deanshi: Is this handled during the map after the `match(_, _)` and the `Tuple.to_list transformation`? Even if it is a tuple?

lib/signs/utilities/messages.ex
deanshi: Moving all of the calls into realtime.ex (recent departures too). Also did a little renaming of scheduled if that's okay.

lib/signs/utilities/messages.ex
PaulJKim: Ah yes you're right, that is handled with the mapping

test/signs/realtime_test.exs
PaulJKim: I don't think this should affect how the early AM feature works with headways unless I missed something in the task description or a discussion? Headways in the early AM only start showing when we're one [lower headway bound] away from the first scheduled departure.

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +287,34 @@ defmodule Signs.Utilities.Messages do
...
+         first_scheduled_departure,
+         last_scheduled_departure,
+         recent_departures,
+         service_ended
+       ) do
+    last_actual_departure =
+      if service_ended do
+        recent_departures
+        |> Enum.max_by(fn {_key, datetime} -> datetime end, fn -> {nil, nil} end)
+        |> elem(1)
```
PaulJKim, line 301: Hmm so this does give use the "last actual departure" relative to a point in time, but I don't think it tells us whether or not that most recent departure was flagged as a last trip. It's possible that even after the last scheduled trip there are multiple trains that run before the "actual last trip" finally passes through a station.

PaulJKim: Hmm valid point. So if service has ended, that means that **a** last trip has departed from the station recently. It _could_ be possible that an extraneous trip passes through that is not necessarily marked as a last trip (I think likely due to human error if anything), but I think your point stands that if service is considered ended, then using the most recent departure should be fine!

lib/engine/scheduled_headways.ex
```diff
@@ -85,6 +85,13 @@ defmodule Engine.ScheduledHeadways do
     |> min_time()
   end
 
+  @callback get_last_scheduled_departure([binary]) :: nil | DateTime.t()
+  def get_last_scheduled_departure(stop_ids) do
+    get_first_last_departures(stop_ids)
+    |> Enum.map(&elem(&1, 1))
+    |> min_time()
```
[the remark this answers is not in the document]
deanshi: yep, that's a super good point. I think for last_scheduled should return the max rather than the min. Let me take a look at the logic.

lib/signs/realtime.ex
```diff
@@ -131,11 +131,22 @@ defmodule Signs.Realtime do
...
+          SourceConfig.sign_stop_ids(config)
+        )
+      end)
+
     prev_predictions_lookup =
       for prediction <- sign.prev_predictions, into: %{} do
         {prediction_key(prediction), prediction}
       end
 
+    recent_departures =
```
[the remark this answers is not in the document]
deanshi: Yep, to grab the latest of the recent departures coming back from the recent_departures call, in cases where we have multiple departures (i.e. red trunk). Further down, we check which of the red_trunk departures is even older, and we pick that one. 

I can simplify the logic here to make it more straightforward if we want.

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +289,41 @@ defmodule Signs.Utilities.Messages do
...
+         current_time,
+         first_scheduled_departure,
+         last_scheduled_departure,
+         recent_departures,
+         service_ended
+       ) do
+    last_actual_departure =
+      if service_ended do
+        recent_departures
+        |> Map.values()
```
PaulJKim, line 302: I think Map.values() returns the values based on the order of the keys (`trip_ids` in this case) which I'm not confident would guarantee ordering for us, considering we often get added trips with ids like `ADDED-12345`. I think we may need to sort the values before acting on the ordering of the departures

lib/signs/realtime.ex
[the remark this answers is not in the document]
deanshi: Okay - so the slightly new flow that I'm proposing now is that we get the recent departures, we flat_map to only the departure dates and we get the _oldest_ date. With the oldest date, we will pass it through to `messages.ex` and that will be the value we use to calculate the buffer. 

In cases where we have two configs to look at, it will still map to the oldest date for each of those configs, and then further down the line still be used for the calculation for each of the configs. 

Does that work or is there something I'm misunderstanding/missing?

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +289,41 @@ defmodule Signs.Utilities.Messages do
...
+        |> case do
+          [] ->
+            nil
+
+          [only_departure] ->
+            only_departure
+
+          [first_departure, second_departure] ->
```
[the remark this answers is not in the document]
deanshi: This shouldn't be a problem anymore - we're just using the oldest departure time passed via map_config from realtime.ex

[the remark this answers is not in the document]
PaulJKim: Ah right, that makes sense!

lib/signs/realtime.ex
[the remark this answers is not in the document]
deanshi: Yep, sorry that's what I meant (the most _recent_ departure). 

The flat_map portion I think is what threw me off, by making it into the tuple pair, but having looked at it I think it makes sense to me now.

All tests are still running properly as well, which is a good indication! And I swapped over to Stream, I think I initially used Enum for testing purposes, but then I forgot to change it back 🤦

lib/signs/utilities/messages.ex
```diff
@@ -40,24 +45,50 @@ defmodule Signs.Utilities.Messages do
...
         else
-          [{sign.source_config, predictions, alert_status, scheduled, service_status}]
+          [
+            {sign.source_config, predictions, alert_status, first_scheduled_departures,
+             last_scheduled_departures, recent_departures, service_status}
+          ]
         end
-        |> Enum.map(fn {config, predictions, alert_status, scheduled, service_status} ->
+        |> Enum.map(fn {config, predictions, alert_status, first_scheduled_departures,
+                        last_scheduled_departures, recent_departures, service_status} ->
```
PaulJKim, line 60: Bit of a nitpick, but could we rename `recent_departures` to something like `most_recent_departure` or something to be a little clearer on what that value is?

PaulJKim (approved):
One nit, but otherwise looks good! Thanks for working through this!

Also before we merge it main, I'd consider letting the logic run overnight and checking the results via Splunk. You can deploy your branch to dev-green to do so

panentheos (approved):
Looks like Paul's review is no longer valid, so tossing mine on here.

PR #935, opened 2025-06-23
Overnight period logic remerge with fixes to crashes

deanshi:
#### Summary of changes

**Asana Ticket:** [Fix and re-merge PA/ESS overnight mode](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210604891580461?focus=true)

Same as https://github.com/mbta/realtime_signs/pull/919 for the first commit. Second commit fixes crashes. 
- Pulled out the tuple zipping to before we check the sign config. This allows us to check if custom text should properly be in the overnight period, but otherwise just passes it normally to the regular handling.
- Adds some extra functions to guard for a `nil` `last_scheduled_departure` and `first_scheduled_departure`. The thought here being we don't know if we're in the overnight period if there isn't a schedule, but this maybe could be relaxed to just be `first_scheduled_departure`?
- Fixed the buffer time to go forward 30 minutes instead of back
- Added/modified tests to account for these three test cases, using a mezz sign, nil departures and checking explicitly for the thirty minute buffer. 

#### Reviewer Checklist

- [x] If `signs.json` was changed, there is a follow-up PR to `signs_ui` to update its dependency on this project

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -6,6 +6,7 @@ defmodule Signs.Utilities.Messages do
 
   @early_am_start ~T[03:29:00]
   @early_am_buffer -40
+  @overnight_buffer 30
 
   @spec get_messages(
           Signs.Realtime.predictions(),
@@ -15,6 +16,8 @@ defmodule Signs.Utilities.Messages do
           Engine.Alerts.Fetcher.stop_status()
           | {Engine.Alerts.Fetcher.stop_status(), Engine.Alerts.Fetcher.stop_status()},
           DateTime.t() | {DateTime.t(), DateTime.t()},
+          DateTime.t() | {DateTime.t(), DateTime.t()},
+          DateTime.t() | {DateTime.t(), DateTime.t()},
           boolean() | {boolean(), boolean()}
         ) :: [Message.t()]
   def get_messages(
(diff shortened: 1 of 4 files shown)


lib/signs/utilities/messages.ex
```diff
@@ -258,6 +304,73 @@ defmodule Signs.Utilities.Messages do
...
+
+  defp overnight_period?(
+         current_time,
+         first_scheduled_departure,
+         last_scheduled_departure,
+         most_recent_departure,
+         true
+       ) do
+    if Timex.after?(most_recent_departure, last_scheduled_departure) do
+      most_recent_departure
```
panentheos, line 346: I'd assume that if we got a definitive last-trip tag, we'd want to use that time as the end of service, even if it's before the scheduled end, right?

lib/signs/utilities/messages.ex
```diff
@@ -23,41 +26,84 @@ defmodule Signs.Utilities.Messages do
         sign_config,
         current_time,
         alert_status,
-        scheduled,
+        first_scheduled_departures,
+        last_scheduled_departures,
+        most_recent_departure,
         service_status
       ) do
+    config =
```
panentheos, line 34: This is good pulled out this way, so we only have to do the transformation once.

lib/signs/utilities/messages.ex
```diff
@@ -258,6 +304,73 @@ defmodule Signs.Utilities.Messages do
...
+
+  defp calculate_overnight_period(
+         last_time_to_check,
+         current_time,
+         first_scheduled_departure
+       ) do
+    # Overnight, schedules might switch to the next day
+    # so we just check if we're before the AM period, or if we're after the last time we need to check
+    Timex.after?(current_time, Timex.shift(last_time_to_check, minutes: @overnight_buffer)) ||
+      before_early_am_threshold?(current_time, first_scheduled_departure)
```
panentheos, line 371: Just checking my understanding: The API automatically flips schedules over at 3AM, so we'd transition from the first part of the `||` being `true` to the second part being `true` atomically, and so we'll remain in the overnight period across that boundary. Right?

PR #938, opened 2025-07-02
Silver Line Headways

deanshi:
#### Summary of changes

**Asana Ticket:** [Actual headway mode on Silver Line signs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1207891972773461?focus=true)

Building off of @robbie-sundstrom PR here: https://github.com/mbta/realtime_signs/pull/933
- Added headway mode for Silver Line buses for platform/mezzanine modes
- For headways on the silver line, we'll now show "Silver Line Buses/Every x to y min" and "Buses to [dest]/Every x to y min".
- Mezzanine (currently seems like just to WTC) will show headways separately until both configs are in headway mode in which case we'll display the "Silver Line Buses" messaging (n.b. currently, at least in my dev-green configuration there doesn't seem to be predictions for WTC Mezz sign to begin with. Will look into)
- Chelsea Bridge paginates with Headways when both need to be shown.
- All audios and tts_audios have been updated to use the headway messaging and audios

Screenshot 2025-07-15 at 9 02 00 AM
Screenshot 2025-07-15 at 9 02 42 AM
Screenshot 2025-07-15 at 9 03 34 AM
Screenshot 2025-07-15 at 9 08 50 AM
Screenshot 2025-07-15 at 9 11 57 AM
Screenshot 2025-07-15 at 11 46 18 AM

#### Reviewer Checklist

### DIFF
--- a/lib/signs/bus.ex
+++ b/lib/signs/bus.ex
@@ -1,6 +1,7 @@
 defmodule Signs.Bus do
   use GenServer
   require Logger
+  alias Signs.Utilities.Headways
 
   @line_max 18
   @var_max 45
@@ -9,7 +10,7 @@ defmodule Signs.Bus do
   @drawbridge_minutes_spanish "152"
   @drawbridge_soon_spanish "157"
   @alert_levels [:station_closure, :suspension_closed_station]
-  @sl_waterfront_route_ids MapSet.new(["741", "742", "743", "746"])
+  @width 24
 
   @enforce_keys [
     :id,
@@ -95,7 +96,9 @@ defmodule Signs.Bus do
               route_id: Map.fetch!(source, "route_id"),
               direction_id: Map.fetch!(source, "direction_id")
             }
-          end
+          end,
+        headway_group: Map.get(config, "headway_group"),
+        headway_direction_name: Map.get(config, "headway_direction_name")
       }
     end
   end
@@ -162,7 +165,7 @@ defmodule Signs.Bus do
     # Compute new sign text and audio
     {[top, bottom], audios, tts_audios} =
       cond do
-        config_off?(config, all_route_ids) ->
+        config_off?(config) ->
           {_messages = ["", ""], _audios = [], _tts_audios = []}
 
         match?({:static_text, _}, config) ->
@@ -196,6 +199,7 @@ defmodule Signs.Bus do
             state
           )
 
+        # Only reach here if top_configs and bottom_configs are defined, indicating a mezz. sign
         true ->
           mezzanine_mode_content(
             predictions_lookup,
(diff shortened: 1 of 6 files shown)


lib/signs/bus.ex
```diff
@@ -434,34 +455,57 @@ defmodule Signs.Bus do
...
     if !Enum.any?(top_content ++ bottom_content, &match?({:predictions, _}, &1)) &&
-         all_stop_ids(state)
+         stop_ids
          |> Enum.all?(fn stop_id ->
            Map.get(stop_alerts_lookup, stop_id) in @alert_levels
          end) do
       no_service_content()
     else
+      silver_line_headway =
+        combine_silver_line_headway_messages(
```
deanshi, line 470: Taken from how we handle rail. 

If the ranges are the same, we combine them into one headway message, otherwise we don't use it.

lib/signs/bus.ex
```diff
@@ -868,6 +1016,36 @@ defmodule Signs.Bus do
...
+    destination_message_id =
+      case destination do
+        :chelsea ->
+          "133"
+
+        :south_station ->
+          "134"
+
+        _ ->
+          "silver_line"
```
deanshi, line 1043: TBD on ARINC ID, already emailed Andrew with the new audio, just waiting to hear back.

lib/signs/bus.ex
```diff
@@ -434,34 +455,57 @@ defmodule Signs.Bus do
...
-        for content <- [top_content, bottom_content] do
-          Enum.map(content, &format_message(&1, nil, current_time))
-          |> paginate_message()
+        if top_content ++ bottom_content == [] do
+          format_message(silver_line_headway, current_time)
+        else
+          for {content, config} <- [{top_content, top_configs}, {bottom_content, bottom_configs}] do
+            if content == [] do
+              headway_message_for_configs(config, stop_ids, current_time)
+              |> format_message(current_time)
```
panentheos, line 482: I don't think this will handle the case where one direction has predictions and the other is showing headways. We should have a test case for that anyway, so writing that will be a good starting point. `format_message` is expected to produce single-line content, so we should keep it that way and have it call `Message.to_single_line` for headways. We'll also need to port/borrow some of the sizing and flipping logic from `Utilities.Messages.render_messages`, because single-line headway messages are too long for the top line.

lib/signs/bus.ex
```diff
@@ -492,6 +536,102 @@ defmodule Signs.Bus do
...
+        message =
+          Headways.headway_message_sl(
+            group,
+            silver_line_destination(name),
+            stop_ids,
+            current_time
+          )
+
+        case message do
+          nil -> {:headway, []}
```
panentheos, line 557: Instead of generating an empty message here and then having to handle that downstream, what about reframing things a bit? If we built this logic into `configs_content`, then we wouldn't have to call it repeatedly, and we could simply not produce a message at all if we don't have the relevant headway data.

lib/signs/utilities/headways.ex
```diff
@@ -21,9 +25,33 @@ defmodule Signs.Utilities.Headways do
...
+  @spec headway_message_sl(
+          String.t(),
+          PaEss.destination() | nil,
+          [String.t()] | nil,
+          DateTime.t()
+        ) ::
+          Message.t() | nil
+  def headway_message_sl(_headway_group, _headway_destination, stop_ids, _current_time)
+      when stop_ids == nil or stop_ids == [],
+      do: nil
```
panentheos, line 38: nit: This may be overly cautious, since I don't think `stop_ids` should ever be `nil`, and the downstream code looks like it will accept and empty array and return `false`.

panentheos (approved):
See a few small cleanup comments, then looks good! Thanks for all the iteration on this. We should do a bit of full-stack validation in a non-prod environment, then this should be good to go.

deanshi:
Deployed to dev-green, and everything is looking like it's working properly. I'll let it sit overnight for now.

deanshi:
Alright - going to merge this. Checking [Splunk](https://mbta.splunkcloud.com/en-US/app/search/search?display.page.search.mode=fast&dispatch.sample_ratio=1&q=search%20index%3D%22realtime-signs-dev-green%22%20failure%20OR%20terminating%20OR%20error&earliest=-24h%40h&latest=%40h&workload_pool=&display.prefs.events.offset=40&sid=1752840113.6896) data, it seems like we're only getting errors dealing with some prediction stuff that was happening before. 

Also checking Screenplay, seems like headways are responding properly with Prediction Suppression, so I think we're looking good.

PR #939, opened 2025-07-08
send single zone to scully

panentheos:
#### Summary of changes

**Asana Ticket:** [Middleware: Batch encode sign and device data](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210475118985495?focus=true)

This switches to the new API format for Scully, and sends a single zone list including the station code. Also fixes a few incorrect SCU references, noticed during hardware encoding.

This change will need to be deployed simultaneously with mbta/scully#98

### DIFF
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -41,7 +41,7 @@ defmodule PaEss.Updater do
         scu_id,
         {:background, scu_id,
          %{
-           visual_zones: [text_zone],
+           zones: ["#{pa_ess_loc}-#{text_zone}"],
            visual_data: visual,
            expiration: 180,
            tag: tag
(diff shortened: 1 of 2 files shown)


lib/pa_ess/updater.ex
```diff
@@ -41,7 +41,7 @@ defmodule PaEss.Updater do
         scu_id,
         {:background, scu_id,
          %{
-           visual_zones: [text_zone],
+           zones: ["#{pa_ess_loc}-#{text_zone}"],
```
digitalcora, line 44: I kinda prefer the API of the legacy code path where we only have to know the "station code"/`pa_ess_loc` and not also the SCU ID — since the latter can be derived from the former, being required to provide both pieces of information seems redundant. Though I guess nothing would stop us from providing a list of station-zones that span multiple SCUs, and then we'd have to add code to "split" the message and send it to all of them... this isn't a concrete suggestion, then, more wondering if there are opportunities for cleanup here, at the moment when it's easiest to do so.

digitalcora (approved):

PR #957, opened 2025-08-25
feat: Mattapan route-level alerts should affect all stops on route

robbie-sundstrom:
#### Summary of changes

**Asana Ticket:** [PA/ESS: revise handling for route level disruptions](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210850028765294?focus=true)

When we get an alert with an `InformedEntity` that does not have an associated stop_id, we now check if the route this is associated with is Mattapan. If it is, we generate a `stop_status` associated with each `stop_id` on the route.

Note: Since this should only practically impact the Mattapan line, I specifically check for the route being Mattapan. However, this could very easily be expanded to account for route-level disruptions on other lines.

#### Reviewer Checklist


robbie-sundstrom:
@digitalcora that was my initial inclination as well, but it does add some complications with the branching lines. Added parsing of both Red and green line, and it's actually not as messy as I first thought it would be.

lib/engine/alerts/api_fetcher.ex
```diff
@@ -101,18 +101,46 @@ defmodule Engine.Alerts.ApiFetcher do
...
   end
 
+  @spec stop_ids_for_route(String.t(), StationConfig.t()) :: [String.t()]
+  defp stop_ids_for_route(route_id, station_config) do
+    route_ids =
+      case route_id do
+        "Green-E" ->
+          ["Green-E", "Green-E-glx"]
+
+        "Green" ->
```
digitalcora, line 131: This will never be the value of the `route` field of an informed entity, because it is not a route ID that exists. If multiple Green Line branches were affected there would have to be one entity with `route: "Green-B"`, a second entity with `route: "Green-C"`, etc.

digitalcora: I'm not sure what you mean by restructuring? This seems like it should fit right into the existing structure you have.

```elixir
"Green-B" -> ~w[...]
"Green-C" -> ~w[...]
"Green-D" -> ~w[...]
"Green-E" -> ~w[Green-E Green-E-glx]
```

robbie-sundstrom: Oh I misunderstood your explanation! Yes that would fit in perfectly if they're all in the same informed entity. I read that as each branch would have its own entity. Thanks for pointing this out

robbie-sundstrom: In my haste to respond, I did not type my thoughts correctly. Edited my previous comment 😅

digitalcora: Ah, okay, I am un-confused. And yes I'm pretty sure this would all be within the same alert.

robbie-sundstrom: I would still like to add a unit test case for the Green Line, but I've updated the logic to include alerts that affect the entire Green Line. Also restructured things so it's a bit easier to understand I think.

lib/engine/alerts/station_config.ex
```diff
@@ -8,7 +8,8 @@ defmodule Engine.Alerts.StationConfig do
   @type t :: %__MODULE__{
           stop_to_station: %{(stop_id :: String.t()) => station()},
           station_to_stops: %{station() => [stop_id :: String.t()]},
-          station_neighbors: %{station() => [station()]}
+          station_neighbors: %{station() => [station()]},
+          route_to_stops: %{String.t() => [stop_id :: String.t()]}
```
digitalcora, line 12: I think we should be careful with our terminology here — what this calls a "route" is not actually a route, it's a sort of "segment" identifier. Sometimes the segment and route are 1:1 (Orange) but sometimes not (Red-Trunk/Ashmont/Braintree).

lib/engine/alerts/api_fetcher.ex
```diff
@@ -101,16 +101,54 @@ defmodule Engine.Alerts.ApiFetcher do
...
+        # If an informed entity does not contain a stop,
+        # then the entire route and all associated stop_ids are affected
+        %{"route" => route}, {stops, routes} ->
+          {stops, [route | routes]}
+      end)
+
+    # When the full green line is affected by an alert, it contains an informed entity
+    # for each branch. In this case, we want to include all stops on GL trunk as well.
+    all_affected_routes =
+      case(includes_all_gl_branches?(all_routes)) do
```
digitalcora, line 122: **refactoring suggestion:** This should definitely work but I think the logic has grown a bit unwieldy. Fundamentally we want to map a set of routes represented in an alert's informed entities to a set of "segments" (whatever we call those identifiers). We could express this idea as a static data structure like:

```elixir
@route_segments [
  {~w[Blue], ~w[Blue]},
  {~w[Green-B], ~w[Green-B]},
  {~w[Green-C], ~w[Green-C]},
  {~w[Green-D], ~w[Green-D]},
  {~w[Green-E], ~w[Green-E Green-E-glx]},
  {~w[Green-B Green-C Green-D Green-E], ~w[Green-Trunk]},
  {~w[Orange], ~w[Orange]},
  {~w[Mattapan], ~w[Mattapan]},
  {~w[Red], ~w[Red-Trunk Red-Ashmont Red-Braintree]}
]
```

Basically, for each of these entries, if _all of_ the route IDs in the first element are present in the informed entities, then add the segment IDs in the second element to our list of segments (then get all the stop IDs for those segments).

For that matter I'm not sure why we couldn't restructure `priv/stops.json` itself to include the information about which routes a segment "belongs to" from the get-go — this data is intimately tied to those segments, and would have to change whenever it changed, so it seems fragile to locate it in a different module. Depends how much of an appetite for refactoring you have!

robbie-sundstrom: Moving this into the  `stops.json` makes a lot of sense to me because as you said, the mapping of segments to larger routes is inherently tied to how we've set up the data there. Working on this now with a structure of 
``` 
{
  "Green-Trunk": 
  {
    "segments": ["Green-B, Green-C", "Green-D", "Green-E"],    
    "stations": [
      {"station": "G Kenmore eastbound", "stop_ids": ["70150", "71151"]},
      {"station": "G Hynes eastbound", "stop_ids": ["70152"]},
      ...
    ]
  },
  ...
}

robbie-sundstrom: Have reworked a bunch of wording and added some comments to try to make things clearer here. Each `segment` maps to a list of `route_ids` that compose that `segment`.

digitalcora (approved):
Approving and merging this in @robbie-sundstrom's absence.

PR #1014, opened 2026-04-06
pass audio items to Scully for decoding

panentheos:
#### Summary of changes

**Asana Ticket:** [Bandwidth-related timeouts with large PA messages](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213791022736120?focus=true)

This modifies the active message sending code to pass the audio item list to Scully using the new API option there, so that resolution of audio data can be done on the SCU (and cached).

This must be deployed _after_ mbta/scully#203


digitalcora (approved):
✨

PR #1020, opened 2026-05-08
add pauses in audio readouts

panentheos:
#### Summary of changes

**Asana Ticket:** [Add breaks around key elements of PA/ESS audio](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214236074987522?focus=true)

This adds pauses to the TTS outputs in order to emphasize the important parts of the messages. Includes a couple more cases not specified in the original ticket, but that follow the same intent:
* Train descriptions in non-revenue announcements, boarding announcements, first train messages, etc.
* Platform letters in "track change" announcements.
* Route names in long-form bus sign readouts.

The test changes provide a nice record of the effective behavior changes.

### DIFF
--- a/lib/content/audio/approaching.ex
+++ b/lib/content/audio/approaching.ex
@@ -94,7 +94,7 @@ defmodule Content.Audio.Approaching do
           do: PaEss.Utilities.four_cars_text(),
           else: " Please stand back from the platform edge."
 
-      "Attention passengers: The next #{train} is now approaching#{platform}#{new_cars}.#{followup}#{crowding}"
+      "Attention passengers: The next#{train}; is now approaching#{platform}#{new_cars}.#{followup}#{crowding}"
     end
 
     defp platform_token(:ashmont), do: :on_the_ashmont_platform
--- a/lib/content/audio/approaching.ex
+++ b/lib/content/audio/approaching.ex
@@ -67,8 +67,8 @@ defmodule Content.Audio.Approaching do
       train = PaEss.Utilities.train_description(audio.destination, audio.route_id, :visual)
       approaching = if audio.four_cars?, do: "now approaching", else: "is now approaching"
       crowding = PaEss.Utilities.crowding_text(audio.crowding_description)
-      platform = platform_string(audio.platform)
-      new_cars = new_cars_string(audio.new_cars? and not audio.four_cars?)
+      platform = if(s = platform_string(audio.platform), do: " #{s}")
+      new_cars = if(s = new_cars_string(audio), do: ", #{s}")
 
       followup =
         if audio.four_cars?,
@@ -87,24 +87,30 @@ defmodule Content.Audio.Approaching do
       train = Utilities.train_description(audio.destination, audio.route_id)
       crowding = PaEss.Utilities.crowding_text(audio.crowding_description)
       platform = platform_string(audio.platform)
-      new_cars = new_cars_string(audio.new_cars? and not audio.four_cars?)
+      new_cars = new_cars_string(audio)
 
       followup =
         if audio.four_cars?,
           do: PaEss.Utilities.four_cars_text(),
           else: " Please stand back from the platform edge."
 
-      "Attention passengers: The next#{train}; is now approaching#{platform}#{new_cars}.#{followup}#{crowding}"
+      main_sentence =
+        ["Attention passengers: The next, #{train}", "is now approaching", platform, new_cars]
+        |> PaEss.Utilities.tts_sentence()
+
+      "#{main_sentence}#{followup}#{crowding}"
     end
 
     defp platform_token(:ashmont), do: :on_the_ashmont_platform
     defp platform_token(:braintree), do: :on_the_braintree_platform
 
-    defp platform_string(:ashmont), do: " on the Ashmont platform"
-    defp platform_string(:braintree), do: " on the Braintree platform"
-    defp platform_string(_), do: ""
+    defp platform_string(:ashmont), do: "on the Ashmont platform"
+    defp platform_string(:braintree), do: "on the Braintree platform"
+    defp platform_string(_), do: nil
 
-    defp new_cars_string(true), do: ", with all new Red Line cars"
-    defp new_cars_string(false), do: ""
+    defp new_cars_string(audio) when audio.new_cars? and not audio.four_cars?,
+      do: "with all new Red Line cars"
+
+    defp new_cars_string(_), do: nil
   end
 end
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -195,9 +195,7 @@ defmodule PaEss.Updater do
   defp format_audio(text), do: %{type: "tts", text: ssml(text), voice_id: "Matthew"}
 
   defp ssml(text) do
-    content = xml_escape(text)
-
-    ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{content}</prosody></amazon:effect></speak>)
+    ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{xml_escape(text)}</prosody></amazon:effect></speak>)
   end
 
   defp create_tag() do
--- a/lib/pa_ess/updater.ex
+++ b/lib/pa_ess/updater.ex
@@ -195,7 +195,7 @@ defmodule PaEss.Updater do
   defp format_audio(text), do: %{type: "tts", text: ssml(text), voice_id: "Matthew"}
 
   defp ssml(text) do
-    content = xml_escape(text) |> String.replace("[pause]", "<break />")
+    content = xml_escape(text)
 
     ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{content}</prosody></amazon:effect></speak>)
   end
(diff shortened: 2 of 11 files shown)


lib/pa_ess/updater.ex
```diff
@@ -195,7 +195,9 @@ defmodule PaEss.Updater do
   defp format_audio(text), do: %{type: "tts", text: ssml(text), voice_id: "Matthew"}
 
   defp ssml(text) do
-    ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{xml_escape(text)}</prosody></amazon:effect></speak>)
+    content = xml_escape(text) |> String.replace("[pause]", "<break />")
```
digitalcora: Thinking ahead to implementing SSML support in Screenplay PSAs, does this interact at all with that, or is it a separate code path?

test/signs/bus_test.exs
```diff
@@ -513,8 +513,8 @@ defmodule Signs.BusTest do
         ],
         [
           {[
-             "The next route 14 bus to Wakefield Ave arrives in 2 minutes.",
-             "The following route 14 bus to Wakefield Ave arrives in 11 minutes."
+             "The next [pause] route 14 [pause] bus to [pause] Wakefield Ave [pause] arrives in [pause] 2 [pause] minutes.",
+             "The following [pause] route 14 [pause] bus to [pause] Wakefield Ave [pause] arrives in [pause] 11 [pause] minutes."
            ], nil},
           {"The Chelsea Street bridge is raised. We expect this to last for at least 4 more minutes. SL3 buses may be delayed, detoured, or turned back.",
```
digitalcora, line 519: Should the number of minutes in this message also be set off with breaks?

test/signs/realtime_test.exs
```diff
@@ -225,7 +225,7 @@ defmodule Signs.RealtimeTest do
       expect_audios(
         [{:canned, {"112", spaced(["501", "787", "920", "933", "21014", "925"]), :audio_visual}}],
         [
-          {"The next Southbound train does not take passengers. Please stand back from the platform edge.",
+          {"The next [pause] Southbound [pause] train does not take passengers. Please stand back from the platform edge.",
```
digitalcora: Subjective, but I think this sounds better with an additional pause after "train". Without it, the delivery makes "train" sound like the beginning of a new sentence ("Train does not take passengers.") This was also the ARINC status quo, iirc.

Applies to "The next X train..."

* does not take passengers
* is stopped Y stops away
* is now boarding
* is now approaching
* arrives in Y minutes
* departs in Y minutes

Perhaps it would make sense to universally follow the word "train" with a break if it is part of a "train description" that is not for a Green Line branch (since there the different message format already results in a break before the "train status", after the destination).

test/signs/realtime_test.exs
```diff
@@ -540,7 +540,7 @@ defmodule Signs.RealtimeTest do
 
       expect_audios(
         [{:canned, {"115", spaced(["501", "4100", "864", "533", "641", "5008", "534"]), :audio}}],
-        [{"The next Mattapan train is stopped 8 stops away.", nil}]
+        [{"The next [pause] Mattapan [pause] train is stopped 8 stops away.", nil}]
```
digitalcora: With the added breaks everywhere else, feels like we're missing a break around the number of stops-away.

```xml
The next <break /> Mattapan <break /> train <break />
is stopped <break /> 8 <break /> stops away.
```

test/signs/realtime_test.exs
```diff
@@ -1223,7 +1230,8 @@ defmodule Signs.RealtimeTest do
         ],
         [
           {"Southbound trains every 11 to 13 minutes.", nil},
-          {"The next Alewife train arrives in 4 minutes on the Ashmont platform.", nil}
+          {"The next [pause] Alewife [pause] train arrives in [pause] 4 [pause] minutes on the Ashmont platform.",
```
digitalcora: As above, with the added breaks elsewhere, I think the absence of a break in the phrase "minutes on the Ashmont platform" (or "minutes on track 1") sounds odd here. Maybe:

```xml
The next <break /> Alewife <break /> train <break />
arrives in <break /> 4 <break /> minutes <break /> on the Ashmont platform.
```

This is exposing a notable weakness of the `<break />` approach, which is that the engine seems to treat it as identical to a comma — but patterns of commas like this basically don't occur in everyday speech, and the resulting tone pattern for "long" sentences like this is really unnatural, almost uncomfortable to listen to.

Experimentally, inserting semicolons (which have the same pause length as commas/break tags) seems to work as a "tone break", varying the delivery just a bit:

```xml
The next <break /> Alewife <break /> train;
arrives in <break /> 4 <break /> minutes;
on the Ashmont platform.
```

I'm hesitant to rely on this effect since punctuation is not an explicit instruction to the engine and ultimately it's free to do whatever it wants... but unfortunately we have very few explicit instructions available to us (and _they_ don't behave 100% consistently either), so maybe it's worth trying this. In that case we could also try using a semicolon for the suggested break after "train" in my other comment.

And, supposing we use punctuation to control breaks anyway, I wonder if there really _is_ any difference between `<break />` and `,`... the latter would be much less obtrusive in code than `[pause]` and would not require us to have this separate "token language" that we replace with XML.

lib/pa_ess/updater.ex
panentheos: All TTS currently passes through here, so obviously we'll need to rethink the escaping strategy if we're supporting SSML input directly.

test/signs/bus_test.exs
panentheos: I'm less convinced about the bridge message. Since it's already a fairly long sentence, it sounds odd to me with a gap in there.

digitalcora: Fair enough, since nothing else here has a pause, it might sound odd with a pause in that one place.

panentheos:
I reworked the code based on the suggestions, using commas to emphasize key words, and semicolons to break things up into "phrases".

lib/content/audio/approaching.ex
```diff
@@ -94,7 +94,7 @@ defmodule Content.Audio.Approaching do
           do: PaEss.Utilities.four_cars_text(),
           else: " Please stand back from the platform edge."
 
-      "Attention passengers: The next #{train} is now approaching#{platform}#{new_cars}.#{followup}#{crowding}"
+      "Attention passengers: The next#{train}; is now approaching#{platform}#{new_cars}.#{followup}#{crowding}"
```
digitalcora: I can't find a corresponding test case but it seems like this particular message, versus e.g. the "arrives in X minutes" one, may be missing a break in the event that there is a platform: `The next, Alewife, train; is now approaching; on the Ashmont platform; with all new Red Line cars.` (semicolons vs. commas doesn't seem to actually make a difference in this case, but it feels like we might as well use semicolons due to these being different conceptual "sections" of the message)

lib/pa_ess/utilities.ex
```diff
@@ -514,13 +514,13 @@ defmodule PaEss.Utilities do
 
     case {route_text, av} do
       {nil, :audio} ->
-        "[pause] #{destination_text} [pause] train"
+        ", #{destination_text}, train"
```
digitalcora: I wonder if it would make more sense to move this leading `, ` into the higher-level messages that embed the train description, so they are "visible" there as part of the message structure, don't depend on the message having a particular spacing (or lack of spacing) before them, and both the leading and trailing breaks around the train description are located in the embedding message (as opposed to _only_ the trailing break, if one exists, which is an odd asymmetry).

test/signs/realtime_test.exs
```diff
@@ -1070,9 +1067,8 @@ defmodule Signs.RealtimeTest do
           {:canned, {"115", spaced(["667", "4016", "864", "503", "504", "5002", "505"]), :audio}}
         ],
         [
-          {"The next [pause] Ashmont [pause] train arrives in [pause] 1 [pause] minute.", nil},
-          {"The following [pause] Ashmont [pause] train arrives in [pause] 2 [pause] minutes.",
-           nil}
+          {"The next, Ashmont, train; arrives in, 1, minute.", nil},
```
digitalcora, line 1110: The delivery of this common message format is sounding better with the semicolon rather than all commas/breaks, imo; it's good that this approach also seems to improve things here in addition to addressing the cases with a _lot_ of breaks.

lib/pa_ess/updater.ex
```diff
@@ -195,7 +195,9 @@ defmodule PaEss.Updater do
   defp format_audio(text), do: %{type: "tts", text: ssml(text), voice_id: "Matthew"}
 
   defp ssml(text) do
-    ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{xml_escape(text)}</prosody></amazon:effect></speak>)
+    content = xml_escape(text)
+
+    ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{content}</prosody></amazon:effect></speak>)
```
digitalcora: This change could be entirely reverted, since it's now a no-op.

digitalcora (approved):
Looking (well, sounding) good! A few trailing commas... I mean, comments.

PR #1021, opened 2026-05-15
fix: suppress pa messages overnight 

rwaskiewicz:
#### Summary of changes

**Asana Ticket:** [Suppressing Overnight PA/ESS readouts](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214305709282852?focus=true)

**Note to Reviewers**

Additional tests for this fix are needed, but I'd like to make sure I'm on the correct route here.

This PR is split into three commits. They are (hopefully) self-describing, but there are a handful of cases where I've put comments directly inline with the source for one reason or another.

Generally, I would have split this into 2 PRs - one containing the refactor + typo fixes, and one with the actual fix. I've included them both in this PR to demonstrate the intent of the refactor. Feel free to review the first commit, and if that's simply the wrong direction to take this, let me know/reject the PR and we'll iterate from there.

**Changes**

Pull the predictions, alert status, departure information, etc. that was
previously calculated in `get_messages/9` into a separate function,
grouping said information into a struct called `SignPredictionInfo`.
This separation allows the calculated information to be used by other
processes, other than `get_message/9`. As a result of wrapping this info
in a struct, the arity of `get_message` has been reduced from 9 to 2.

Simple unit tests have been created in `messages_test.exs` to smoke test
the refactor. While not these do not fully cover the derivation of
messages, they are a step in the right direction

Suppress non-emergency messages such that they do not play overnight.

Fix minor typos found in comments when going through relevant parts of
the codebase.

#### Reviewer Checklist

### DIFF
--- a/lib/engine/pa_messages.ex
+++ b/lib/engine/pa_messages.ex
@@ -48,10 +48,14 @@ defmodule Engine.PaMessages do
     {:noreply, state}
   end
 
+  @spec select_pa_messages([PaMessages.PaMessage.t()], state()) ::
+          {[PaMessages.PaMessage.t()], state()}
   defp select_pa_messages(pa_messages, state) do
     now = DateTime.utc_now()
 
-    Enum.flat_map_reduce(pa_messages, state, fn message, state ->
+    pa_messages
+    |> Enum.reject(&is_overnight_and_non_emergency?/1)
+    |> Enum.flat_map_reduce(state, fn message, state ->
       last_sent = Map.get(state.pa_messages_last_sent, message.id, DateTime.from_unix!(0))
 
       if DateTime.diff(DateTime.utc_now(), last_sent, :millisecond) >= message.interval_in_ms do
@@ -60,6 +64,17 @@ defmodule Engine.PaMessages do
         {[], state}
       end
     end)
+  end
+
+  @spec is_overnight_and_non_emergency?(PaMessage.t()) :: boolean()
+  defp is_overnight_and_non_emergency?(pa_message) do
+    # emergencies will have a priority of 1 or less
+    Enum.all?(pa_message.sign_ids, &is_sign_in_overnight_period?(&1)) && pa_message.priority >= 2
+  end
+
+  @spec is_sign_in_overnight_period?(String.t()) :: boolean
+  defp is_sign_in_overnight_period?(sign_id) do
+    GenServer.call(:"Signs/#{sign_id}", {:overnight_status})
   end
 
   defp play_pa_messages(pa_messages) do
--- a/lib/engine/pa_messages.ex
+++ b/lib/engine/pa_messages.ex
@@ -53,9 +53,7 @@ defmodule Engine.PaMessages do
   defp select_pa_messages(pa_messages, state) do
     now = DateTime.utc_now()
 
-    pa_messages
-    |> Enum.reject(&is_overnight_and_non_emergency?/1)
-    |> Enum.flat_map_reduce(state, fn message, state ->
+    Enum.flat_map_reduce(pa_messages, state, fn message, state ->
       last_sent = Map.get(state.pa_messages_last_sent, message.id, DateTime.from_unix!(0))
 
       if DateTime.diff(DateTime.utc_now(), last_sent, :millisecond) >= message.interval_in_ms do
@@ -64,17 +62,6 @@ defmodule Engine.PaMessages do
         {[], state}
       end
     end)
-  end
-
-  @spec is_overnight_and_non_emergency?(PaMessage.t()) :: boolean()
-  defp is_overnight_and_non_emergency?(pa_message) do
-    # emergencies will have a priority of 1 or less
-    Enum.all?(pa_message.sign_ids, &is_sign_in_overnight_period?(&1)) && pa_message.priority >= 2
-  end
-
-  @spec is_sign_in_overnight_period?(String.t()) :: boolean
-  defp is_sign_in_overnight_period?(sign_id) do
-    GenServer.call(:"Signs/#{sign_id}", {:overnight_status})
   end
 
   defp play_pa_messages(pa_messages) do
--- a/lib/signs/realtime.ex
+++ b/lib/signs/realtime.ex
@@ -115,14 +115,14 @@ defmodule Signs.Realtime do
   end
 
   def handle_call({:play_pa_message, pa_message}, _from, sign) do
-    {sign, should_play?} = Signs.Utilities.Audio.handle_pa_message_play(pa_message, sign)
-
     sign_in_overnight_period? =
       derive_sign_context(sign)
       |> Signs.Utilities.Messages.flatten_sign_context(sign)
       |> Signs.Utilities.Messages.in_overnight_period?()
 
-    should_play? = should_play? && !sign_in_overnight_period?
+    {sign, should_play?} =
+      Signs.Utilities.Audio.handle_pa_message_play(pa_message, sign, sign_in_overnight_period?)
+
     {:reply, {sign, should_play?}, sign}
   end
 
--- a/lib/signs/realtime.ex
+++ b/lib/signs/realtime.ex
@@ -119,6 +119,16 @@ defmodule Signs.Realtime do
     {:reply, {sign, should_play?}, sign}
   end
 
+  @spec handle_call({:overnight_status}, {pid(), term()}, t()) :: {:reply, boolean(), t()}
+  def handle_call({:overnight_status}, _from, sign) do
+    sign_in_overnight_period =
+      derive_sign_prediction_info(sign)
+      |> Signs.Utilities.Messages.flatten_sign_prediction_info(sign)
+      |> Signs.Utilities.Messages.in_overnight_period?()
+
+    {:reply, sign_in_overnight_period, sign}
+  end
+
   # TODO(now): Call out in review I will move this, it makes the diff easier in GH
   @spec derive_sign_prediction_info(t()) :: Signs.Utilities.SignPredictionInfo.t()
   defp derive_sign_prediction_info(sign) do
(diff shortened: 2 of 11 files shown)


lib/engine/pa_messages.ex
```diff
@@ -62,6 +66,17 @@ defmodule Engine.PaMessages do
     end)
   end
 
+  @spec is_overnight_and_non_emergency?(PaMessage.t()) :: boolean()
+  defp is_overnight_and_non_emergency?(pa_message) do
+    # emergencies will have a priority of 1 or less
+    Enum.all?(pa_message.sign_ids, &is_sign_in_overnight_period?(&1)) && pa_message.priority >= 2
```
rwaskiewicz: While we get the `message_type` back from Screenplay, we don't store it in the message struct when we deserialize that JSON response. I don't know if that was an intentional design choice or not, which is why the priority level is used here. Happy to add `message_type` here if there are cases where there are messages of high priority that aren't emergencies, or some other reason where this isn't the proper solution.

lib/signs/utilities/messages.ex
```diff
@@ -8,67 +8,36 @@ defmodule Signs.Utilities.Messages do
   @early_am_buffer -40
   @overnight_buffer 30
 
-  @spec get_messages(
-          Signs.Realtime.predictions(),
+  @type flattened_sign_prediction_info :: {
```
rwaskiewicz: Open to other names here (and in the functions that are related to it) - this is the best I could think of 😅

lib/signs/utilities/sign_prediction_info.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Signs.Utilities.SignPredictionInfo do
```
rwaskiewicz: Open to other names here (and in the functions that are related to it)

lib/signs/utilities/messages.ex
```diff
@@ -8,67 +8,36 @@ defmodule Signs.Utilities.Messages do
...
-        last_scheduled_departures,
-        most_recent_departure,
-        service_status
+        sign_prediction_info = %Signs.Utilities.SignPredictionInfo{
+          current_time: current_time,
+          sign_config: sign_config
+        }
       ) do
     config =
-      if match?(%{source_config: {_, _}}, sign) do
```
rwaskiewicz, line 35: This logic is largely abstracted as-is to here - https://github.com/mbta/realtime_signs/pull/1021/changes#diff-bdda4cc40319158f7ed70a48350c960125e1cc22848b0c5f9dccde12c05205beR91

lib/signs/realtime.ex
```diff
@@ -119,7 +119,19 @@ defmodule Signs.Realtime do
...
+  def handle_call({:overnight_status}, _from, sign) do
+    sign_in_overnight_period =
+      derive_sign_prediction_info(sign)
+      |> Signs.Utilities.Messages.flatten_sign_prediction_info(sign)
+      |> Signs.Utilities.Messages.in_overnight_period?()
+
+    {:reply, sign_in_overnight_period, sign}
+  end
+
+  # TODO(now): Call out in review I will move this, it makes the diff easier in GH
```
rwaskiewicz: Like the comment implies, GH's diff algo makes it very clear (IMO) that the existing logic was extracted to a new fn. I plan on moving this below its first usage after the review.

lib/signs/realtime.ex
```diff
@@ -174,17 +186,32 @@ defmodule Signs.Realtime do
...
+      alert_status: alert_status,
+      first_scheduled_departures: first_scheduled_departures,
+      last_scheduled_departures: last_scheduled_departures,
+      recent_departures: recent_departures,
+      service_end_statuses_per_source: service_end_statuses_per_source
+    }
+  end
+
+  def handle_info(:run_loop, sign) do
+    # TODO(now): Call out in review there may be better abstraction/split here
```
rwaskiewicz: I considered adding two structs here, as `SignPredictionInfo` is currently a big heavy struct that wraps all the information we passed to `get_messages`. I _think_ this is slightly better than it was, but still not perfect. However, to minimize the scope of this PR, I'm OK with leaving this as-is (minus the TODO) and iterating on it in the future

lib/engine/pa_messages.ex
```diff
@@ -48,10 +48,14 @@ defmodule Engine.PaMessages do
...
   end
 
+  @spec select_pa_messages([PaMessages.PaMessage.t()], state()) ::
+          {[PaMessages.PaMessage.t()], state()}
   defp select_pa_messages(pa_messages, state) do
     now = DateTime.utc_now()
 
-    Enum.flat_map_reduce(pa_messages, state, fn message, state ->
+    pa_messages
+    |> Enum.reject(&is_overnight_and_non_emergency?/1)
```
rwaskiewicz: Tests for this fix are needed, but I'd like to make sure I'm on the correct route here

lib/signs/utilities/sign_prediction_info.ex
panentheos: I haven't been able to think of a slam-dunk alternative, but given that this does encompass more than just predictions, it might warrant something a bit more general, e.g. `SignContext` or something similar.

lib/signs/utilities/messages.ex
panentheos: Thinking out loud: Ultimately the "flattened" version is the more useful one, since that's what we actually consume in both of these code paths, so maybe it deserves the prominence of being a struct. But that's something we can tackle in a followup, and right now this PR is well scoped around the immediate problem, which is good.

lib/engine/pa_messages.ex
panentheos: The sign ids can span multiple stations, which may have different overnight statuses at a given time. Instead of requiring all signs to be overnight before we suppress, I think we want to move this down as a filter that's applied per-sign. There's already a call to `:play_pa_message` that goes through the sign process and can skip playback, so that would be a convenient place to do this.

panentheos (commented):
The approach here looks good, and is along the lines of what I was envisioning. I called out a few tweaks below, and I'll do a full review once it's marked ready.

test/signs/utilities/messages_test.exs
```diff
@@ -0,0 +1,277 @@
...
+          alert_status: :none,
+          first_scheduled_departures: DateTime.shift(@midnight, hour: -1),
+          last_scheduled_departures: DateTime.shift(@midnight, hour: 1),
+          recent_departures: @midnight,
+          service_end_statuses_per_source: false
+        }
+      }
+    end
+
+    test "returns an empty message when sign is in the overnight period", %{
```
panentheos, line 192: FYI, because of how this app is structured, we've been able to lean heavily on higher level tests that exercise the full `:run_loop`. See `realtime_test.exs` for an extensive suite. They tend to be about as concise as lower level unit tests, while exercising more code in context and being immune to churn from internal refactoring. It's fine to have a few spot checks here, but for more comprehensive validation, we can take advantage of those high level tests (and fill in gaps there if applicable)

lib/engine/pa_messages.ex
rwaskiewicz: > The sign ids can span multiple stations, which may have different overnight statuses at a given time. 

Ah, okay - that's good to know. For my own edification, do you have an example off the top of your head? A sign only "spanning" 1 station was the association I was assuming here 

> There's already a call to :play_pa_message that goes through the sign process and can skip playback, so that would be a convenient place to do this.

Thanks! I think that'll help clean some of the solution-related code up. I'll look into integrating a revised version of this logic there now

lib/signs/utilities/messages.ex
rwaskiewicz: Agreed - I held off there since the flattened version touched a bit more code. If there's bandwidth at the end of the sprint, I'll look into this refactor

lib/engine/pa_messages.ex
rwaskiewicz: Oh, maybe I misunderstood - you're saying a message can have multiple signs, and those signs for that message can span multiple stations. 🤦

panentheos: As an example, PSA messages will usually be scheduled on a whole line or lines at once, so this list may have dozens of sign ids. They get grouped by `pa_ess_loc`, which is roughly equivalent to station-level, in `play_pa_messages`.

panentheos: Oh I see, my original wording was confusing. The sign itself is only at a single station. PA messages often play to many signs at once, so the _list of ids_ can span multiple stations.

rwaskiewicz: Gotcha, that makes sense.

test/signs/utilities/messages_test.exs
rwaskiewicz: Makes sense - the tests I added here and in the new `audio_tests.exs` file were moreso for my own confidence while refactoring. This file a little less so, since I cover the simpler cases in it

lib/signs/utilities/audio.ex
```diff
@@ -199,18 +199,43 @@ defmodule Signs.Utilities.Audio do
...
+  def handle_pa_message_play(pa_message, %Signs.Realtime{} = sign, true)
+      when pa_message.priority >= 2 do
+    log_skipped_pa_message(pa_message, sign)
+    {sign, false}
+  end
+
+  @spec handle_pa_message_play(PaMessages.PaMessage.t(), Signs.Realtime.t(), boolean()) ::
+          {Signs.Realtime.t(), boolean()}
+  def handle_pa_message_play(pa_message, sign, _overnight_mode) do
+    handle_pa_message_play(pa_message, sign)
```
panentheos, line 213: nit: This could be simplified by removing this function clause and adding the unused third parameter to the one below.

panentheos (approved):
See minor comment, and there are a few TODOs to remove, then looks good!

lib/signs/utilities/audio.ex
rwaskiewicz: I held off on doing that, only because this utility is also used for buses as well, and didn't want to change those callsites in `bus.ex` ([example](https://github.com/mbta/realtime_signs/blob/0422ee13c55d4724d683d5456fa73764505e8443/lib/signs/bus.ex#L121)). However, if that's more idiomatic, I'm happy to do so (just LMK, easy enough follow up)

PR #1022, opened 2026-05-18
expand abbreviations in TTS for migrated stations

panentheos:
#### Summary of changes

**Asana Ticket:** [Replace abbreviations in TTS for custom messages at migrated stations](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214901763748036?focus=true)

This replaces abbreviations in custom messages when generating TTS readouts for migrated stations, same as we do for ARINC stations. Also unifies the code paths between rail and bus, so they work the same.

### DIFF
--- a/lib/pa_ess/utilities.ex
+++ b/lib/pa_ess/utilities.ex
@@ -596,7 +596,7 @@ defmodule PaEss.Utilities do
   end
 
   @spec replace_abbreviations(String.t()) :: String.t()
-  def replace_abbreviations(text) when is_binary(text) do
+  defp replace_abbreviations(text) when is_binary(text) do
     Enum.reduce(
       @abbreviation_replacements,
       text,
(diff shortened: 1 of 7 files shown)


lib/pa_ess/utilities.ex
```diff
@@ -606,6 +606,21 @@ defmodule PaEss.Utilities do
...
+    string
+    |> String.replace(@invalid_custom_character, "")
+    |> String.slice(0, if(line == :top, do: 18, else: 24))
+  end
+
+  @spec custom_tts_text(String.t(), String.t()) :: String.t()
+  def custom_tts_text(top, bottom) do
+    "#{validate_custom_string(top, :top)} #{validate_custom_string(bottom, :bottom)}"
+    |> String.trim()
+    |> replace_abbreviations()
```
digitalcora, line 621: This seems to not have any remaining uses outside this module — can be made private? (it still has dedicated test coverage, but you also added test coverage of the relevant behavior in `realtime_test`)

digitalcora (approved):

PR #1025, opened 2026-05-22
refactor sign context

panentheos:
#### Summary of changes

This formalizes the new `SignContext` struct and promotes the "flattened" form into its own `ConfigContext` struct. Most of the message generation code now deals with these structs, which streamlines parameter passing.

This reactor ended up being a prerequisite for the PA message lookahead feature.

Most of the changes are mechanical, except for moving the headway information into the context instead of doing an ad-hoc fetch. All the test changes are adjusting the mocks for the new call order; there should be no user visible changes.

### DIFF
--- a/lib/signs/utilities/messages.ex
+++ b/lib/signs/utilities/messages.ex
@@ -8,36 +8,12 @@ defmodule Signs.Utilities.Messages do
   @early_am_buffer -40
   @overnight_buffer 30
 
-  @type flattened_sign_context :: {
-          Signs.Utilities.SourceConfig.config(),
-          list(Signs.Realtime.predictions()),
-          Signs.Realtime.t(),
-          Engine.Alerts.Fetcher.stop_status(),
-          DateTime.t(),
-          DateTime.t(),
-          DateTime.t(),
-          boolean(),
-          boolean()
-        }
+  alias Signs.Utilities.SignContext
+  alias SignContext.ConfigContext
 
-  @spec get_messages(
-          Signs.Realtime.t(),
-          Signs.Utilities.SignContext.t()
-        ) :: [Message.t()]
-  def get_messages(
-        sign,
-        sign_context = %Signs.Utilities.SignContext{
-          current_time: current_time,
-          sign_config: sign_config
-        }
-      ) do
-    config =
-      flatten_sign_context(
-        sign_context,
-        sign
-      )
-
-    sign_in_overnight_period = in_overnight_period?(config)
+  @spec get_messages(Signs.Realtime.t(), SignContext.t()) :: [Message.t()]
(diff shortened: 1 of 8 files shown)


lib/signs/utilities/messages.ex
```diff
@@ -383,12 +307,20 @@ defmodule Signs.Utilities.Messages do
...
+         %SignContext{sign_config: sign_config} = sign_context,
+         sign
+       ) do
+    alert_status = filter_alert_status(alert_status, sign_config)
     route = Signs.Utilities.SourceConfig.single_route(config)
     transfer_alert? = alert_status in [:suspension_transfer_station, :shuttles_transfer_station]
     union_square? = sign.pa_ess_loc == "GUNS"
 
     cond do
+      overnight_period?(config_context, sign_context) ->
```
rwaskiewicz, line 321: For my own understanding, are we adding this clause as a consequence of the refactor, or is this something we should have added in https://github.com/mbta/realtime_signs/commit/7888c95f7f544e064bab562c12ada0a338a5d6de?

lib/signs/utilities/messages.ex
```diff
@@ -407,13 +339,24 @@ defmodule Signs.Utilities.Messages do
...
 
-  defp service_ended_message(true = _service_ended?, config, false = _sign_in_overnight_period) do
+  defp service_ended_message(
+         %ConfigContext{config: config} = config_context,
+         %SignContext{} = sign_context
+       ) do
     route = Signs.Utilities.SourceConfig.single_route(config)
 
-    %Message.ServiceEnded{destination: config.headway_destination, route: route}
+    if config_context.service_ended? and not in_overnight_period?(sign_context) do
```
rwaskiewicz, line 358: Similar to my previous question, should we have included this in https://github.com/mbta/realtime_signs/pull/1025/changes#r3289766028? Do service ended messages play at a fixed interval for the some or the entire period in which there isn't service?

rwaskiewicz (approved):

lib/signs/utilities/messages.ex
panentheos: This used to be precomputed and passed as a separate parameter, which was handled via the separate function clause below. Now that it's computed on the fly, we need to roll it into the logic explicitly.

lib/signs/utilities/messages.ex
panentheos: This was also rolled in from what used to be a separate parameter. Service ended messages are a bit unusual in that they pay attention to the _sign-level_ overnight status instead of just the config-level one. So if one direction is still running, we'll keep showing "service ended" in the other, but as soon as they both flip over then the sign goes blank.

PR #1027, opened 2026-05-26
play PA messages per-screen

panentheos:
#### Summary of changes

This refactors PA message playback to be done per-sign instead of grouped per-station. This is a prerequisite for being able to drop or defer PA messages based on local factors. Note that this will result in more individual messages being sent to Scully, but after the recent audio caching changes, this should have no significant performance implications.


lib/engine/pa_messages.ex
```diff
@@ -4,85 +4,52 @@ defmodule Engine.PaMessages do
...
+  end
+
   def start_link([]) do
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
   end
 
   @impl true
   def init(_) do
-    # Add some delay to wait for sign processes to come up before sending PA messages
-    Process.send_after(self(), :update, 10000)
```
rwaskiewicz, line 20: For my own understanding, what's changed such that we don't need a 10 second delay to start the update loop and can now use `schedule_update`/it's 1 second delay?

rwaskiewicz (approved):

lib/engine/pa_messages.ex
panentheos: Previously we were proactively sending messages to sign processes, and those processes have their own short sleep at the beginning to avoid bringing down the whole app in response to some transient errors. They wouldn't be able to respond during that sleep, so we waited until well after it was done. Now that it's flipped and the signs are polling the engine (actually just reading the ETS table), there's no dependency there so we can start the loop immediately.

PR #1031, opened 2026-06-01
Revert PA message lookahead and per-sign scheduling

panentheos:
This reverts #1029 and #1027, rolling back the recent changes to PA message logic.


rwaskiewicz (approved):
LGTM - the only two commits between landing #1029 and `main@HEAD` don't have any overlap with these. 🚢

PR #1048, opened 2026-07-10
deps: remove unused `configparser_ex`

digitalcora:
This is an optional dependency of ExAws and only needed to load AWS CLI config files. It appears to be an artifact of when RTS was deployed to on-prem Windows servers, since the dependency is `prod`-only and we don't have the AWS CLI installed in the Docker containers we deploy to ECS today.

Also remove configuration that configured ExAws using `SIGNS_S3_CONFIG_` environment variables in non-prod environments; these are not found in the current `.envrc.template`. Having done this, the entire config block can be deleted, since it is the same as the default value[^1].

[^1]: https://github.com/ex-aws/ex_aws#aws-key-configuration

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -49,10 +49,6 @@ config :realtime_signs,
 config :realtime_signs, RealtimeSignsWeb.Endpoint,
   secret_key_base: "local_secret_key_base_at_least_64_bytes_________________________________"
 
-config :ex_aws,
-  access_key_id: [{:system, "SIGNS_S3_CONFIG_KEY"}, :instance_role],
-  secret_access_key: [{:system, "SIGNS_S3_CONFIG_SECRET"}, :instance_role]
-
 config :logger, utc_log: true
 
 config :logger, :console,
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -5,11 +5,3 @@ config :logger, :console, level: :info
 config :realtime_signs,
   external_config_getter: ExternalConfig.S3,
   restart_fn: &System.restart/0
-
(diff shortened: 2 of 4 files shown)


config/config.exs
```diff
@@ -49,10 +49,6 @@ config :realtime_signs,
 config :realtime_signs, RealtimeSignsWeb.Endpoint,
   secret_key_base: "local_secret_key_base_at_least_64_bytes_________________________________"
 
-config :ex_aws,
-  access_key_id: [{:system, "SIGNS_S3_CONFIG_KEY"}, :instance_role],
```
rwaskiewicz, line 53: Double checking, it looks like the default value you linked to includes a `:pod_identity` atom, which looks like it maps to [this AWS functionality](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html). I'm not familiar with how EKS Pod Identity works, but would it be correct to say that:
1. Falling back to the default value would add `:pod_identity` to our `:ex_aws` config and enable EKS Pod Identity, but
2. Wouldn't use it, since we don't have any association maps that associate an IAM role to the pods running in EKS 
3. And none of it matters, because this service doesn't run in EKS

rwaskiewicz (approved):

config/config.exs
digitalcora: That seems accurate. We've been using the default authentication values for some time without issue in our other ECS apps, at least Screens and Screenplay.

rwaskiewicz: Gotcha, thanks for confirming!

PR #1055, opened 2026-07-20
clean up legacy code paths

panentheos:
#### Summary of changes

**Asana Ticket:** [Remove ARINC specific code from RTS](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216519965021338?focus=true)

This removes the legacy middleware code paths now that all stations have been migrated. Despite the large diff, there should be nothing surprising.

Note: Removes the fallback conditional when parsing custom message TTS strings, since all configs have been updated to follow the new schema now.


config/runtime.exs
```diff
@@ -27,8 +27,6 @@ if config_env() != :test do
     scu_ip_map: System.get_env("SCU_IP_MAP", "{}") |> JSON.decode!(),
     chelsea_bridge_password: System.get_env("CHELSEA_BRIDGE_PASSWORD"),
     chelsea_bridge_username: System.get_env("CHELSEA_BRIDGE_USERNAME"),
-    number_of_http_updaters:
```
rwaskiewicz, line 30: Once this lands/is deployed, does this mean we can remove this field from our Terraform config?

lib/pa_ess/utilities.ex
```diff
@@ -833,429 +590,6 @@ defmodule PaEss.Utilities do
...
-    {"Gallivan Blvd", "881"},
-    {"Brookline Ave", "885"},
-    {"Brookline Village", "886"},
-    {"Cobbs Corner Canton", "887"},
-    {"Linden Sq", "889"}
-  ]
-
-  @spec audio_take(term()) :: String.t()
-  # Tokens ending in underscores are newer Polly-generated clips
-  def audio_take(:eastbound), do: "867"
```
rwaskiewicz, line 923: Non-blocking - it looks like these fn clauses were generated with:
https://github.com/mbta/realtime_signs/blob/ba1f0bc3d4054f80a9d52bbda20048182f0076e5/scripts/one_offs/station_abbreviations.exs#L156-L176 

Since the name is "one off", I assume we don't need to update this script but wanted to verify that

rwaskiewicz (approved):

lib/pa_ess/utilities.ex
panentheos: Yeah I like to commit these scripts so they can be reviewed at the time, but they're not intended to be run again and we'll delete them at some point.

PR #1056, opened 2026-07-20
remove outdated scripts

panentheos:
#### Summary of changes

This removes a couple one-off codegen scripts that are no longer needed, as well as the Latency Analysis script which dealt with legacy log data.


digitalcora (approved):
✂️ ✂️ ✂️

PR #1060, opened 2026-07-29
fix: group messages by scu instead of paess location

rwaskiewicz:
#### Summary of changes

**Asana Ticket:** [PA/ESS: Prevent duplicate PA readouts at shared transfer station zones](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216645320223225?focus=true)

Change the grouping of messages to be by their SCU, rather than their PAESS location identifier.

Prior to this commit, public announcements could play back to back at a station when a section of the physical station space had more than one audio zone, where each audio zone was associated with a unique PA/ESS location. Previously, this was caused by grouping by `pa_ess_loc`, which caused the same message to be sent to the SCU twice, often in quick succession, causing the "back to back" readout.

#### Reviewer Checklist


lib/engine/pa_messages.ex
```diff
@@ -66,20 +66,35 @@ defmodule Engine.PaMessages do
...
-      |> Enum.group_by(& &1.pa_ess_loc)
-      |> Enum.map(fn {_, [first | _] = signs} ->
-        %{
-          first
-          | audio_zones: Enum.flat_map(signs, & &1.audio_zones) |> Enum.uniq(),
-            id: Enum.map_join(signs, ",", & &1.id)
-        }
-      end)
-      |> Signs.Utilities.Audio.send_audio([message])
+      signs_and_zone_overrides =
```
rwaskiewicz: Very open to suggestions here to make this better. The thought process is that we need generate the zone "overrides" here, because we previously assumed that we'd always be [using a single PA ESS Location ID](https://github.com/mbta/realtime_signs/pull/1060/changes#diff-38dd9d87909f26303fadbadfe5013c800cf2ead6450db49ce7b850c9e158cfd8L113) for rail and bus right before we play the message - without this change we end up creating invalid zones and/or dropping zones to play the message in. 

I added a log message in dev that shows the actual payload sent to SCUs, which looks something like this with this change:
Image

lib/pa_ess/updater.ex
```diff
@@ -50,10 +50,11 @@ defmodule PaEss.Updater do
               [Signs.Realtime.t() | Signs.Bus.t()],
               [Content.Audio.tts_value()],
               integer(),
-              [keyword()]
+              [keyword()],
+              [String.t()]
             ) ::
               :ok
-  def play_message(signs, tts_audios, priority, log_metas) do
+  def play_message(signs, tts_audios, priority, log_metas, zone_overrides) do
```
rwaskiewicz: Interestingly, I had a default value for `zone_overrides` here originally, but the tests for `bus.exs` didn't seem to be picking it up, regardless of how I updated the mocks.

panentheos:
Ah, I see why this is not straightforward as-is. Now that the pre-migration code is gone, there's nothing special about `pa_ess_loc`, and no reason it needs to be separate from the zone letters. What if we first did a small refactor to pre-compute the full Scully zone names on initial load and store those in `text_zone` and `audio_zones`? There are a couple places looking at `pa_ess_loc` or those fields for special behavior, but I think they should be easy to tweak. Then the PA message grouping change could be trivial.

rwaskiewicz:
> Ah, I see why this is not straightforward as-is. Now that the pre-migration code is gone, there's nothing special about `pa_ess_loc`, and no reason it needs to be separate from the zone letters. What if we first did a small refactor to pre-compute the full Scully zone names on initial load and store those in `text_zone` and `audio_zones`? There are a couple places looking at `pa_ess_loc` or those fields for special behavior, but I think they should be easy to tweak. Then the PA message grouping change could be trivial.

Sounds good -  I'll start working on that now

rwaskiewicz:
@panentheos I've repurposed this PR with 2 commits
- https://github.com/mbta/realtime_signs/pull/1060/commits/4e56fc73676a00e8d64dd9e829493aefa06ec897 is the refactor. 
- https://github.com/mbta/realtime_signs/pull/1060/commits/3ab5957e3ffa3ac1eff83a0b3fa1d7da04538af7 groups messages by SCU, which is so wonderfully trivial now 😆 

Here's an example with a debug statement (that I've since removed from this branch) showing we can target multiple zones with different PAESS Locations:
Screenshot 2026-07-30 at 15 15 21

Can you take a look when you get a chance?

panentheos (approved):
This is great, nice and clean!

PR #1062, opened 2026-08-03
chore: switch from Hackney to Req

digitalcora:
* Use Req functions and structs for HTTP requests. This generally involves providing `headers` as part of options rather than a separate argument (and these can use a more concise keyword-based format), replacing `HTTPoison.Response` with `Req.Response` and swapping `status_code` for `status`, and swapping `recv_timeout` for `receive_timeout`.

* Remove `timeout` option on requests; this was not a valid option in Hackney and so did nothing.

* Adjust various code to use Req's map format for response headers.

* Remove manual JSON decoding of response data; this is automatic by default in Req. Also remove some defensive code paths for the very unlikely case of a request for a JSON resource having a successful status code, but not returning valid JSON.

* Remove manual JSON and form-data *encoding* of request data; Req can do this automatically, including setting appropriate request headers, by providing the body as a `json` or `form` option rather than `body`.

* Use the built-in compression support (`compressed: true`) for V3 API requests.

* Remove some leftover "fake requests" in tests that were specific to the ARINC code paths (removed in ea1fb25).

* Remove `Engine.NetworkCheck`. This was a workaround for a bug in Hackney 1.x (potentially only in combination with Windows networking, since this problem dates to when the app ran on Windows) where it could get stuck in a bad state but not cause any outright crashes, requiring us to "manually" detect this and restart the whole app.

* Remove the "ARINC" connection pool used for SCU requests. Finch does not support using the same pool for requests to different scheme/host/port combinations (...nor would this really make sense, on reflection, since the point of a pool is to speed up requests to *the same* S/H/P by leaving connections to it open; unclear what this actually did in Hackney). Allowing each SCU its own auto-created Finch pool does not impact memory use since connections are started lazily.

* Convert existing `pool_info` metrics logging to look at Finch pools. The logging format is kept the same in case anything depends on this.

**Asana Ticket:** https://app.asana.com/1/15492006741476/project/1185117109217413/task/1200085512957624?focus=true

#### Reviewer Checklist

- [x] Meets ticket's acceptance criteria (technically)
- [x] This branch was deployed to dev-green and is currently running with no unexpected increase in warnings, and no errors or crashes (compare on Splunk: [dev-green](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-dev-green%20%22%5Berror%5D%22%20OR%20%22%5Bwarning%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840107.3874236) vs. [prod](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drealtime-signs-prod%20%22%5Berror%5D%22%20OR%20%22%5Bwarn%5D%22%20OR%20%22CRASH%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-4h%40m&latest=now&sid=1545840137.3874305))


lib/engine/health.ex
```diff
@@ -61,43 +40,23 @@ defmodule Engine.Health do
...
-  end
-
-  defp log_hackney_pools do
-    Enum.each(
-      @hackney_pools,
-      fn pool ->
-        stats = :hackney_pool.get_stats(pool)
-
-        Logger.info(
-          "event=pool_info name=#{pool} pool_max=#{stats[:max]} in_use_count=#{stats[:in_use_count]} free_count=#{stats[:free_count]} queue_count=#{stats[:queue_count]}"
```
deanshi, line 76: Just making sure - will any of these changes cause any underlying logging differences with our Splunk alerts/dashboards that we'll need to update? (maybe not here specifically except the omission of `queue_count`).

I took a quick scan of our Dashboards and Alerts and it looked okay but not sure if I missed something.

deanshi (approved):
Looks good, seems straightforward especially with the details you provided in the description. Just one small question!

lib/engine/health.ex
digitalcora: I don't think we are meaningfully changing anything here with regards to logging. All the changed or removed logs are relatively unstructured errors/warnings for specific failures that we wouldn't have built dashboards or alerts around. Plus this specific log which does remove `queue_count`, but that's unavoidable as Finch doesn't provide such a metric.

